PR #4, opened 2020-12-09
terraform suite

emesika:
This is minimal suite with DC/Cluster/Host and base networking on which we will test OCP Installer on top of Terraform Provider to ensure that there are no regression cause by latest merged PRs


mwperina:
Please push through gerrit and abandon this PR

sandrobonazzola:
pushed to gerrit

PR #6, opened 2022-02-10
HE: Apply OpenScap profile on rhel8

almusil:


ost_utils/pytest/fixtures/he.py
```diff
@@ -152,6 +152,17 @@ def he_engine_answer_file_storage_snippet(
         raise RuntimeError(f'Unknown ost_he_storage_domain_type {ost_he_storage_domain_type}')
 
 
+@pytest.fixture(scope="session")
+def he_engine_answer_file_openscap_profile_snippet(ost_images_distro):
+    if ost_images_distro == 'rhel8':
+        return (
+            'OVEHOSTED_VM/applyOpenScapProfile=bool:True\n'
+            'OVEHOSTED_VM/OpenScapProfileName=str:stig\n'
```
almusil, line 160: I don't like that we need to put the short there, because the `/root/ost_images_openscap_profile` contains the full name, so we need some matching to have it working also with PCI DSS.

didib: I think I understand what you say, not sure. Generally speaking, 'stig' should probably be the content of some new fixture created one way or another, and we should have code writing this fixture to ost_images_openscap_profile, and also use this fixture here, in the line above. If you meant something else, please clarify.

almusil:
We still need https://github.com/oVirt/ovirt-ansible-collection/pull/426 marking as draft till then.

ost_utils/pytest/fixtures/he.py
```diff
@@ -152,6 +155,16 @@ def he_engine_answer_file_storage_snippet(
         raise RuntimeError(f'Unknown ost_he_storage_domain_type {ost_he_storage_domain_type}')
 
 
+@pytest.fixture(scope="session")
+def he_engine_answer_file_openscap_profile_snippet(ansible_host0):
+    profile = ansible_host0.shell('cat /root/ost_images_openscap_profile')['stdout']
+    if len(profile) > 0:
```
tinez, line 161: Just making sure - there's no newline character if we don't use any profile?

ost_utils/pytest/fixtures/he.py
```diff
@@ -152,6 +152,14 @@ def he_engine_answer_file_storage_snippet(
         raise RuntimeError(f'Unknown ost_he_storage_domain_type {ost_he_storage_domain_type}')
 
 
+@pytest.fixture(scope="session")
+def he_engine_answer_file_openscap_profile_snippet(ost_images_distro):
+    if ost_images_distro == 'rhel8':
```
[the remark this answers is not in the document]
tinez: You are right, every host should have the same profile

ost_utils/pytest/fixtures/he.py
almusil: There shouldn't any at all. The build is doing `echo -n "%OPENSCAP_PROFILE%" > /root/ost_images_openscap_profile` so it should be fine.

common/deploy-scripts/setup_first_he_host.sh
```diff
@@ -96,9 +96,9 @@ copy_dependencies() {
     dest: /var/tmp/lm_sensors.rpm
   with_fileglob:
     - "/var/tmp/lm_sensors-*"
-- name: TODO create empty OpenSCAP profile for now
+- name: Copy OpenSCAP profile name
   copy:
-    content: ""
+    src: /root/ost_images_openscap_profile
```
michalskrivanek, line 101: probably better/safer if it goes to the block below

lagofy.sh
```diff
@@ -162,6 +162,8 @@ ost_init() {
     echo "Suite: $SUITE, distro: $OST_IMAGES_DISTRO, deployment dir: $PREFIX, images:"
     . common/helpers/ost-images.sh
 
+    [[ "$OST_IMAGES_DISTRO" == "rhel8" ]] && OST_IMAGES_NODE="$OST_IMAGES_HOST_INSTALLED"
```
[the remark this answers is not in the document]
almusil: Yeah that was there for testing

almusil:
This is finally working.

PR #8, opened 2022-02-11
Exit with a message when SELinux is disabled

emesika:
chcon is used without checking if SELinux is not disabled.
This cause a message of the form :
"chcon: can't apply partial context to unlabeled file ..."

This patch adds a check for SELinux enforcing status and exists
with a proper message if SELinux is disabled.

Signed-of-by: Eli Mesika <emesika@redhat.com>

### DIFF
--- a/lagofy.sh
+++ b/lagofy.sh
@@ -170,7 +170,7 @@ ost_init() {
     mkdir "$PREFIX"
     mkdir "$PREFIX/logs"
     mkdir "$PREFIX/images"
-    chcon -t svirt_image_t "$PREFIX/images"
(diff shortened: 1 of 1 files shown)


almusil:
This actually breaks OST inside containers, if the message is the only problem what about adding if that checks selinux status?

tinez (commented):
Not sure how is this change supposed to work. IIUC correctly `chcon` call exits with non-0 code if selinux is disabled? If so, then we should rather have something like:

```
[[ "$(getenforce)" == "Enforcing" ]] && chcon ...
```

almusil:
AFAIK selinux is always presenting itself as disabled in container. To check if you are inside container something there is `/run/.containerenv` for Podman and `/run/.dockerenv` for Docker.

michalskrivanek:
i think we can safely close this

PR #9, opened 2022-02-11
package_mgmt: Allow 'el8' packages in rhel8 runs

tinez:
We don't really expect d/s packages to be built on GitHub,
but one exception to this rule is the OCP installer needed
by Terraform suite, that is tested only on rhel8.


michalskrivanek (approved):
/ost

michalskrivanek:
oh..ost is not configured to ost....that needs to be fixed!:)

michalskrivanek:
> oh..ost is not configured to ost....that needs to be fixed!:)

#10 should do the trick, so let's see...

PR #15, opened 2022-02-15
ovirtlib: ansiblelib - fix mutable default value

erav:
With a mutable default value a function doesn't behave idempotently on
consecutive invocations.

Change-Id: I372bbc14459ae07cda200366a50dcc11dd9e4c7f

### DIFF
--- a/network-suite-master/ovirtlib/ansiblelib.py
+++ b/network-suite-master/ovirtlib/ansiblelib.py
@@ -14,11 +14,11 @@ class AnsibleExecutionFailure(Exception):
 
 
 class Playbook(object):
-    def __init__(self, playbook, extra_vars={}):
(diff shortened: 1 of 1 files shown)


erav:
verified: passes on network-suite [1]
[1] https://rhv-devops-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/ds-ost-baremetal_manual/28378/

hbraha (approved):
+1, interesting, do you know what is the root cause for this change of behavior? 
should we open a bug for it?

erav:
> +1, interesting, do you know what is the root cause for this change of behavior? should we open a bug for it?

this is python behaviour. need to investigate what they say about it

hbraha (approved):

tinez (approved):

PR #16, opened 2022-02-15
add support for ost-images CI

michalskrivanek:
this will build a custom set of ost-images and then run
basic-suite-master on them

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -11,6 +11,9 @@
   - initial step - paste first and second part together. You must provide IP= and RHEL= strings at the top, and in Build Environment use secret file OST_APP_PRIVATE_KEY_FILENAME
   - inject variables from "vars"
   - conditinal step(multiple) if build is still succeeding at this point
+    - for ost-images runs only
+        - build specific ost-image first when CUSTOM_OST_IMAGES is set to a string "build-me-some". Will be filled in by build-image job for the respective ost-images refspec from PR
+        - inject vars again, overriding OST_REFSPEC to empty (to run stock basic-suite-master code) and CUSTOM_OST_IMAGES set to the resulting repo from build-image
     - call out the actual OST job with CUSTOM_REPOS=${PR_URL} and OST_IMAGE=${OST_IMAGE} set
     - processing step - paste second part only, and make sure it gets executed regardless the OST job result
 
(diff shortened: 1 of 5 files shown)


michalskrivanek:
jenkins changes already applied, not fully verified yet

.github/README.md
```diff
@@ -11,6 +11,9 @@
   - initial step - paste first and second part together. You must provide IP= and RHEL= strings at the top, and in Build Environment use secret file OST_APP_PRIVATE_KEY_FILENAME
   - inject variables from "vars"
   - conditinal step(multiple) if build is still succeeding at this point
+    - for ost-images runs only
```
tinez, line 14: The indentation here is wrong, if the two points below were supposed to be indented let's end `for ost-images runs only` with `:`

tinez (approved):

lagofy.sh
```diff
@@ -60,6 +60,7 @@ ost_destroy() {
         ) 9>/tmp/ost.lock
     fi
     [[ -s "$PREFIX/sshd_pid" ]] && { echo "killing IPv6 sshd proxy"; kill $(cat "$PREFIX/sshd_pid"); }
+    [[ -d "$OST_REPO_ROOT/custom-ost-images" ]] && { echo "remove custom ost images"; rm -rf "$OST_REPO_ROOT/custom-ost-images"; }
```
tinez, line 63: I don't see the `--justdb` option used anywhere. Don't we need to uninstall those images when destroying?

common/helpers/ost-images.sh
```diff
@@ -1,15 +1,30 @@
 #!/bin/bash -xe
 
-OST_IMAGES_DIR=${OST_IMAGES_DIR:=/usr/share/ost-images}
+if [[ -n "$CUSTOM_OST_IMAGES_REPO" ]]; then
+  echo "Installing custom ost-images from $CUSTOM_OST_IMAGES_REPO"
+  sudo dnf -y --installroot="$OST_REPO_ROOT/custom-ost-images" --releasever=/  install bash
```
tinez, line 5: Will it work even if the directory doesn't exist?

lagofy.sh
michalskrivanek: no, --installroot is basically a regular full chroot, that's why we need to add bash first too

common/helpers/ost-images.sh
michalskrivanek: yes

PR #18, opened 2022-02-16
002_bootstrap: move engine backup before image tests

tinez:
We're still having problems with 'test_upload_cirros_image'
and 'test_create_cirros_template' tests - they fail randomly.
The suggested workaround for now is to wait more before trying
to work with images, so we're moving engine backup test (which
takes ~3 mins to execute) to happen before the problematic tests.


tinez:
Replaced with https://github.com/oVirt/ovirt-system-tests/pull/21

PR #21, opened 2022-02-17
002_bootstrap: wait some more before uploading

bennyz:


tinez:
Let's give it a try for a couple of days. If we'll still get failures for uploading images we'll need to revert it

PR #22, opened 2022-02-20
Revert "basic: Remove time sync"

michalskrivanek:
This reverts commit 8a0261005bff8a3f001dd5ac716f362f679e3b53.

breaks openscap DISA STIG test "Ensure Chrony is only configured with
the server directive"


michalskrivanek:
we need permanent conf change instead

PR #24, opened 2022-02-21
ost_utils: vmconsole - add timeout on read

erav:
Sometimes the vmconsole read hangs and so does the whole job. To avoid
leaving the job hanging forever a timeout will abort it if a read hangs
for more than 15 seconds.

Change-Id: Ibb835297e57b22b20d1331063711b2cad5cad58e

### DIFF
--- a/ost_utils/vmconsole.py
+++ b/ost_utils/vmconsole.py
@@ -9,6 +9,7 @@ import ipaddress
 import logging
 import os
 import pty
+import signal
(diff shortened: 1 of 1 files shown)


erav:
passed as expected on network-suite [1]
timed out on read on he-basic-suite [2]

[1] ds-ost-baremetal_manual/29089
[2] ds-ost-baremetal_manual/29090

PR #27, opened 2022-02-22
Adding tr-suite-master trigger

emesika:
Adds a tr-suite-master trigger file in order to run OST on
tr-suite-master suite using the passed PR URL and distro

Signed-off-by: Eli Mesika <emesika@rehat.com>


tinez:
This should be a PR for a different project

PR #28, opened 2022-02-22
basic: 098 - wait for provider sync completion event

erav:
Engine periodically invokes a sync request to fetch updates from ovn
provider. In test_use_ovn_provider the sync is disabled as part of the
test but if a sync request has not completed before the disable then a
race situation is created which causes failures at various stages in the
test.
To avoid such a race, a wait on the completion of the sync action is
added after disabling it. Either the sync has completed before the wait
or the wait should catch it as 10 seconds is long enough for the whole
round trip of the sync request.

Change-Id: I68dfd4512ccd5de2c1e72b0f638a10b101b86652


basic-suite-master/test-scenarios/test_098_ovirt_provider_ovn.py
```diff
@@ -373,41 +374,42 @@ def test_use_ovn_provider(engine_api, engine_ip_url, engine_full_username, engin
...
-        ovirt_network_id = _get_ovirt_network(engine_api, datacenter_id, NETWORK_1)
-        _add_network_to_cluster(engine_api, datacenter_id, ovirt_network_id)
-        _hotplug_network_to_vm(engine_api, VM0_NAME, NETWORK_1, IFACE_NAME)
-        _remove_iface_from_vm(engine_api, VM0_NAME, IFACE_NAME)
-        _remove_network_from_ovirt(engine_api, datacenter_id, ovirt_network_id)
-
-        _delete_port(token_id, engine_ip_url, port1_id)
-        _delete_subnet(token_id, engine_ip_url, subnet1_id)
-        _delete_network(token_id, engine_ip_url, network1_id)
+        with engine_utils.wait_for_event(engine, 217, 10):  # PROVIDER_SYNCHRONIZED_PERFORMED
```
almusil, line 377: For this reason there were two events added `PROVIDER_SYNCHRONIZATION_STARTED` and `PROVIDER_SYNCHRONIZATION_ENDED`, but I guess this is good enough if it works, worst case it adds 10 seconds.

almusil (approved):

basic-suite-master/test-scenarios/test_098_ovirt_provider_ovn.py
erav: AFAICS on the latest failure engine.log, the _PERFORMED event comes after the _ENDED event and is the last event from the command before releasing the lock.

PR #29, opened 2022-02-23
update the readme document for github

galitf:


michalskrivanek:
> /ost

fyi iiuc with the current workflow code it won't work you since you're oVirt membership is private

tinez (approved):

PR #31, opened 2022-02-23
he: Delay the custom repo check

acernek:
Engine is deployed later than other guests, so we delay the cusom repos
check until it's up.

### DIFF
--- a/ost_utils/pytest/fixtures/deployment.py
+++ b/ost_utils/pytest/fixtures/deployment.py
@@ -96,6 +96,7 @@ def deploy(
     ansible_vms_to_deploy,
     ansible_hosts,
     deploy_scripts,
+    deploy_hosted_engine,
     root_dir,
     working_dir,
     request,
@@ -141,7 +142,7 @@ def deploy(
             'dnf upgrade --nogpgcheck -y --disableplugin versionlock -x ovirt-release-master,ovirt-release-master-tested,ovirt-engine-appliance,rhvm-appliance,ovirt-node-ng-image-update,redhat-virtualization-host-image-update'
         )
         # check if packages from custom repos were used
-        if not request.config.getoption('--skip-custom-repos-check'):
+        if not request.config.getoption('--skip-custom-repos-check') and not deploy_hosted_engine:
             package_mgmt.check_installed_packages(ansible_vms_to_deploy)
 
     # report package versions
--- a/he-basic-suite-master/test-scenarios/test_001_he_deploy.py
+++ b/he-basic-suite-master/test-scenarios/test_001_he_deploy.py
@@ -6,6 +6,10 @@
 
 import os
 
+import pytest
+
+from ost_utils.deployment_utils import package_mgmt
+
 
 def test_run_dig_loop(
     suite_dir,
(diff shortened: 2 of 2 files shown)


acernek:
@tinez @didib please review

michalskrivanek:
i don't quite like the numerous differences/duplication we have because of HE...
it's fine to have the check as a "regular" testcase, but then let's use the same on for all suites...

acernek:
> i don't quite like the numerous differences/duplication we have because of HE... it's fine to have the check as a "regular" testcase, but then let's use the same code in all suites...

ok, but then there could be a problem with the network suite, since we don't number the test modules there (and they're supposed to be independent)

ost_utils/pytest/fixtures/deployment.py
```diff
@@ -141,7 +143,10 @@ def deploy(
             'dnf upgrade --nogpgcheck -y --disableplugin versionlock -x ovirt-release-master,ovirt-release-master-tested,ovirt-engine-appliance,rhvm-appliance,ovirt-node-ng-image-update,redhat-virtualization-host-image-update'
         )
         # check if packages from custom repos were used
-        if not request.config.getoption('--skip-custom-repos-check'):
+        if (
+            not request.config.getoption('--skip-custom-repos-check')
+            and ost_dc_name == engine_object_names.TEST_DC_NAME
```
tinez, line 148: We have a `deploy_hosted_engine` fixture that would be a better choice for the condition here

ost_utils/pytest/fixtures/deployment.py
```diff
@@ -141,7 +143,10 @@ def deploy(
...
+        ):
             package_mgmt.check_installed_packages(ansible_vms_to_deploy)
```
tinez, line 150: Once we're done with this patch I think we can have a look at removing `ansible_vms_to_deploy` completely and just use `ansible_all`. AFAIR the former was there to omit HE VM from the deployment process as it couldn't succeed. Nowadays the HE VM is added dynamically to the inventory, so we most probably don't need that workaround anymore.

tinez:
> > i don't quite like the numerous differences/duplication we have because of HE... it's fine to have the check as a "regular" testcase, but then let's use the same code in all suites...
> 
> ok, but then there could be a problem with the network suite, since we don't number the test modules there (and they're supposed to be independent)

Yes, unfortunately until network suite is "unordered" we have to do things like these. We can improve in subsequent patches.

ost_utils/pytest/fixtures/deployment.py
acernek: Done

ost_utils/pytest/fixtures/deployment.py
acernek: Will try to keep that in mind

michalskrivanek:
I would then rather number it. it's not helpful anyway as it practically always run in same order.
Or even just run an extra pytest session for that one file/tc.

tinez:
> I would then rather number it. it's not helpful anyway as it practically always run in same order. Or even just run an extra pytest session for that one file/tc.

This needs to be discussed with Eitan. For now let's have this PR in as it's an improvement.

tinez (approved):

PR #32, opened 2022-02-23
chrony: Use single server instead of pool

acernek:
Let's make openscap report happy.

### DIFF
--- a/ost_utils/pytest/fixtures/deployment.py
+++ b/ost_utils/pytest/fixtures/deployment.py
@@ -158,5 +158,9 @@ def deploy(
     # setup sar stat utility
     set_sar_interval()
 
+    # setup chrony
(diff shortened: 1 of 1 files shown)


michalskrivanek:
lgtm, does it work?

acernek:
It does fix the openscap fail, but I'm a little bit worried it might lead to bigger time differences between hosts (this way it chooses one server from the pool at random) - unfortunately I will probably not have a time to check the time differences until next week. @tinez

michalskrivanek:
why not adding a new server line with bogus ip/name?

acernek:
If I read it correctly, the openscap test also checks for pools (and fails if it finds one), so I think we have to replace it.

michalskrivanek:
I don't think we should go ahead with this. I prefer to keep DHCP-pushed configuration than losing time sync. In loaded environment the time soon goes off without proper NTP.

I think it's perfectly ok to ignore rules that do not have remediation and need to be manually resolved.

PR #33, opened 2022-02-23
Change cluster firmware to UEFI

nyoxi:
A necessary prerequisite here is that the VM memory is increased otherwise GRUB fails to boot the OS.

NB: Similar memory limit is in `he-basic-suite-master` but it seems we don't run the VMs that are created there. Do we want to change that limit too to keep it in sync?


michalskrivanek:
lgtm

michalskrivanek:
I've merged the #61 for now. It should make HE working, next we need to do CL 4.7 change...

nyoxi:
> @nyoxi any update?

Yes, I have updated the patch. But I see you have already "fixed" the problem in #82 in a different fashion.

PR #38, opened 2022-02-23
setup_host.sh: delete setup for sasl2/qemu.conf

hbraha:
the setup for /etc/sasl2/qemu.conf already as part of the host
deploy[1].
therefore we can remove it from setup_host.sh

[1] https://github.com/oVirt/ovirt-engine/blob/master/packaging/ansible-runner-service-project/project/roles/ovirt-host-setup-vnc-sasl/tasks/main.yml#L8


michalskrivanek:
It's not part of host deploy. That's why it's duplicated. 
@mz-pdm supposedly has a todo to fix this. If not, feel free to fix yourself, but until it's actually called in host deploy we cannot drop it from here

mz-pdm:
> It's not part of host deploy. That's why it's duplicated. @mz-pdm supposedly has a todo to fix this.

I don't think so, feel free to do what's needed.

PR #43, opened 2022-02-28
github: trigger checks on push

erav:
Trigger checks on every push of new code to a PR, instead of only
on PR creation and with manual comment creation. This should be quicker
and more efficient.

Change-Id: I68500b0226a6ccdf98125f7851682fd5038f0951

### DIFF
--- a/.github/workflows/ost-of-ost.yaml
+++ b/.github/workflows/ost-of-ost.yaml
@@ -1,6 +1,7 @@
 name: OST of OST
 
 on:
+  push:
   issue_comment:
     types: [created]
 


erav:
we might be triggering when the PR is created twice, once because a PR comment is created and once for the push, so might need to add a condition on the issue_comment to not trigger when PR first created. I will check this

michalskrivanek:
push event is going to be ignored due to the job condition anyway.

I would think that sometimes we push a bit too often and since the OST jobs are not being cancelled you can easily overload the system

erav:
> push event is going to be ignored due to the job condition anyway.
> 
> I would think that sometimes we push a bit too often and since the OST jobs are not being cancelled you can easily overload the system

any reason there should be more pushes to github than to gerrit?

tinez:
> I would think that sometimes we push a bit too often and since the OST jobs are not being cancelled you can easily overload the system

I second that - I think it's better to trigger manually with `/ost` for now

PR #44, opened 2022-03-01
vdsm-coverage: fix vdsm coverage broken

hbraha:
it seems that in the middle of the process of stopping vdsmd
systemd might already started stopping supervdsmd which vdsmd require.
this issue cause the vdsm coverage to break.

### DIFF
--- a/ost_utils/coverage/vdsm.py
+++ b/ost_utils/coverage/vdsm.py
@@ -68,12 +68,10 @@ def _stop_vdsm_services(hosts):
     try:
         logging.debug('Stopping VDSM services...')
         hosts.systemd(name='vdsmd', state='stopped')
-        hosts.systemd(name='supervdsmd', state='stopped')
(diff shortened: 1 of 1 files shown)


ost_utils/coverage/vdsm.py
```diff
@@ -66,14 +66,15 @@ def _stop_vdsm_services(hosts):
     coverage.py dump coverage data
     """
     try:
+        # comment out supervdsmd as workaround for broken vdsm coverage
         logging.debug('Stopping VDSM services...')
         hosts.systemd(name='vdsmd', state='stopped')
-        hosts.systemd(name='supervdsmd', state='stopped')
+        # hosts.systemd(name='supervdsmd', state='stopped')
```
tinez, line 72: Let's just drop these two lines and leave the comment out

tinez (approved):

PR #46, opened 2022-03-01
README: update the custom repo URL to github

avlitman:
Since we moved to github we are not using jenkins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,8 +44,12 @@ If you have any other questions, please join [oVirt Development forum / mailing 
 Make sure your machine is set up with `setup_for_ost.sh`
 You can use `ost.sh` for running the complete suite on a concrete ost-images distro
 E.g. `./ost.sh run basic-suite-master el8stream`
-You can use `--custom-repo` for verifying an oVirt patch
-E.g `./ost.sh run basic-suite-master el8stream --custom-repo=https://jenkins.ovirt.org/job/ovirt-dwh_standard-check-patch/****/`
+
+You can use `--custom-repo` for verifying an oVirt patch:
+
+* On your GitHub account go to Settings/Developer settings/Personal access tokens/Generate new token, and give it the **repo** scopes.
+* On your ost machine run `export GITHUB_TOKEN=personal_access_token`
+* Run the tests, e.g `./ost.sh run basic-suite-master el8stream --custom-repo=https://github.com/oVirt/ovirt-engine/pull/pr_number`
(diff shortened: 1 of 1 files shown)


tinez (approved):

tinez (changes requested):
Please also mention the necessity to have the `GITHUB_TOKEN` env variable defined

avlitman:
> Please also mention the necessity to have the `GITHUB_TOKEN` env variable defined

Done.

README.md
```diff
@@ -44,9 +44,12 @@ If you have any other questions, please join [oVirt Development forum / mailing
 Make sure your machine is set up with `setup_for_ost.sh`
 You can use `ost.sh` for running the complete suite on a concrete ost-images distro
 E.g. `./ost.sh run basic-suite-master el8stream`
-You can use `--custom-repo` for verifying an oVirt patch
-E.g `./ost.sh run basic-suite-master el8stream --custom-repo=https://jenkins.ovirt.org/job/ovirt-dwh_standard-check-patch/****/`
-You can pass this option multiple times to test more than 1 build at once.
+
+You can use `--custom-repo` for verifying an oVirt patch:
+
+*On your GitHub account go to Settings/Developer settings/Personal access tokens, and generate the token with appropriate permissions.
```
tinez, line 50: The formatting here doesn't work as you'd expect I think - please have a look at the rich diff

tinez: re: `appropriate permissions` personally I always forget which permissions are needed to make it work - I'd really appreciate if they were mentioned explicitly.

tinez (changes requested):

README.md
avlitman: > 

I used the format from lines 72-75, I want it to be a list, I can see a preview somehow ?

avlitman: > re: `appropriate permissions` personally I always forget which permissions are needed to make it work - I'd really appreciate if they were mentioned explicitly.

In my account under 'Select scopes' I selected the 'repo' (include all) and the 'workflow':
Screenshot from 2022-03-01 14-55-08

but really not sure if less/more is needed, can you check your permissions?

avlitman: @tinez So it works with the 'repo'  permissions, you prefer to verify as well ? or I can add that only 'repo' are needed?

tinez (approved):

README.md
tinez: I did verify that too - `repo` seems to suffice

tinez (approved):

PR #52, opened 2022-03-03
github: parse suite and distro from trigger comment

michalskrivanek:
similar to what we had on gerrit, this workflow can trigger different
suites and distros. The new agument "comment" takes the github comment
body in form "/ost[ <suite>[ <distro>]]".

### DIFF
--- a/.github/workflows/ost.yaml
+++ b/.github/workflows/ost.yaml
@@ -1,18 +1,15 @@
 #
-# This is a reusable worflow to be triggered from other projects or manually
+# This is a reusable worflow to be triggered from other projects or manually.
+# See example-trigger.yaml for how to use it in other projects.
+# If you call this via a different trigger than on: issue_comment you can use the optional parameter:
+#  comment - full comment body, e.g. "/ost he-basic-suite-master el9stream"
 #
 name: OST trigger
 
 on:
   workflow_call:
     inputs:
-      pr_url:
-        required: true
-        type: string
-      suite:
-        required: false
-        type: string
-      distro:
+      comment:
         required: false
         type: string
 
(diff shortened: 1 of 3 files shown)


michalskrivanek:
AFAIK no one is really using the previous arguments yet, so it's safe to change the "API"

.github/workflows/ost.yaml
```diff
@@ -39,13 +34,16 @@ jobs:
...
           OST_APP_ID: 164117
           PR_URL: "${{ github.event.inputs.pr_url }}${{ inputs.pr_url }}"
-          SUITE: "${{ github.event.inputs.suite }}${{ inputs.suite }}"
-          DISTRO: "${{ github.event.inputs.distro }}${{ inputs.distro }}"
+          COMMENT: "${{ github.event.inputs.comment }}${{ inputs.comment }}"
         working-directory: /var/lib/nginx/gh
         run: |
           echo "$PR_URL" | grep '^https://api.github.com/repos/oVirt/'
+          echo "$COMMENT" | egrep '^/ost( [a-z\-]+( [a-z0-9]+)?)?$'
+          SUITE=$(echo "$COMMENT" cut -d " " -f2)
```
almusil, line 42: Missing `|`

tinez (approved):

michalskrivanek:
@almusil @tinez any other comment or can we try it live?
manual dispatch works as expected, not sure about the actual calls

almusil:
Looks ok, I am just a bit concerned about backward compatibility, will those old one still work?

.github/workflows/ost.yaml
```diff
@@ -1,18 +1,14 @@
...
   workflow_call:
     inputs:
-      pr_url:
-        required: true
-        type: string
-      suite:
-        required: false
-        type: string
-      distro:
+      comment:
```
almusil, line 11: Why is this still needed?

.github/workflows/ost.yaml
```diff
@@ -21,29 +17,32 @@ on:
...
+      )
     steps:
       - name: trigger
         env:
           OST_APP_ID: 164117
-          PR_URL: "${{ github.event.inputs.pr_url }}${{ inputs.pr_url }}"
-          SUITE: "${{ github.event.inputs.suite }}${{ inputs.suite }}"
-          DISTRO: "${{ github.event.inputs.distro }}${{ inputs.distro }}"
+          PR_URL: "${{ github.event.issue.pull_request.url }}${{ github.event.inputs.pr_url }}"
+          COMMENT: "${{ github.event.comment.body }}${{ inputs.comment }}${{github.event.inputs.comment}}"
```
almusil, line 39: nit: Missing spaces in `${{github.event.inputs.comment}}`

.github/workflows/ost.yaml
michalskrivanek: it's not, it's optional, see line 3

almusil: Right, but I was asking in general, why it remains here even as optional

almusil (approved):

.github/workflows/ost.yaml
```diff
@@ -1,18 +1,14 @@
 #
-# This is a reusable worflow to be triggered from other projects or manually
+# This is a reusable worflow to be triggered from other projects or manually.
+# If you call this via a different trigger than on: issue_comment you can use the optional parameter:
+#  comment - full comment body, e.g. "/ost basic-suite-master"
```
avlitman, line 4: suggestion: because this workflow can trigger different suites and distros maybe add a distro to the example e.g "/ost basic-suite-master el8stream".

michalskrivanek: sure. but note it's optional, you do not need any parameter unless you use the workflow differently than intended

avlitman: > sure. but note it's optional, you do not need any parameter unless you use the workflow differently than intended

+1 Thanks!

PR #55, opened 2022-03-03
decrease heading sizes

galitf:


tinez:
Re: commit message and PR title - we're not fixing anything, we're just decreasing heading sizes

tinez (changes requested):
See my comments re: commit msg and PR title

tinez (approved):

PR #56, opened 2022-03-04
he: workaround for old appliance

michalskrivanek:
add RHEL repos since we added some new dependencies and we don't have
appliance nightly


michalskrivanek:
this part works, the rest of he suite...hard to say

PR #58, opened 2022-03-07
Allow capture errors when mkdocs build fails

galitf:


tinez (changes requested):
Docs are not linters really. I'd rather see a commit that adds a new github workflow that runs `docs` separately. We don't want it in regular OST runs.

PR #59, opened 2022-03-07
HE: Pass IP Version to deployment script

didib:
This is required now: https://bugzilla.redhat.com/2021975

Change-Id: I2de04907b136400b4916ac5b2daac540d7751c49


common/deploy-scripts/setup_first_he_host.sh
```diff
@@ -70,7 +72,7 @@ dnf_update() {
   dnf:
     name:  "*"
     state: latest
-    enablerepo: "*-baseos,*-appstream"
+    enablerepo: "rhel*-baseos,rhel*-appstream,appstream,ovirt-*testing"
```
michalskrivanek, line 75: please no ovirt repos. ovirt appliance is built nightly, nothing should be needed

michalskrivanek (changes requested):

common/deploy-scripts/setup_first_he_host.sh
didib: done

didib:
This isn't enough to make he-basic pass, but is useful anyway. Verified locally only on cs8

common/deploy-scripts/setup_first_he_host.sh
```diff
@@ -9,6 +9,8 @@ HE_MAC_ADDRESS="$1"
 shift
 HEADDR="$1"
 shift
+IP_VERSION_OPTION="$1"
```
michalskrivanek, line 12: I would rather duplicate the logic in the script than pass/require another parameter. 
Especially since we have that logic here already at line 156

common/deploy-scripts/setup_first_he_host.sh
```diff
@@ -70,7 +72,7 @@ dnf_update() {
   dnf:
     name:  "*"
     state: latest
-    enablerepo: "*-baseos,*-appstream"
+    enablerepo: "rhel*-baseos,rhel*-appstream"
```
michalskrivanek, line 75: makes sense to "enhance" the comment above that we want/need this only on rhv and so it's effective only on rhel8

michalskrivanek (changes requested):

common/deploy-scripts/setup_first_he_host.sh
michalskrivanek: taking this change out into its own PR #65

didib:
Of the three commits currently in this PR, one is merged as #65, another is not needed anymore (pass IP version) because we reverted the patch that required it (#30). Rebasing and keeping the remaining one.

didib:
Passed basic-suite, generated pytest.log looks ok. I think it can be merged.

PR #61, opened 2022-03-14
he, basic: increase VM memory size for UEFI boot

michalskrivanek:
In Cluster 4.7 we now default to UEFI boot, that needs a bit more
memory.


michalskrivanek:
just the memory part of #33 
should fix HE runs

michalskrivanek:
i don't know what's up with the connectivity...eh

PR #62, opened 2022-03-15
Remove test for export button

ljelinkova:
This patch removes tests for the Export button on the vm list view. The reason is that we test only top level action buttons, and the export button has been moved to the kebab menu.


michalskrivanek (approved):

michalskrivanek:
it's so embarrassing that after 3 years we still cannot reliably upload an image :(

PR #66, opened 2022-03-17
setup: Enably deny debug for fapolicyd

almusil:
Anything that is denied by fapolicyd should be
logged to journal when the unit runs with
--debug-deny. This should help to identify
any "Operation not permitted" that might be related
to fapolicyd.

Signed-off-by: Ales Musil <amusil@redhat.com>


michalskrivanek:
@almusil isn't it easier to create during image creation so that it boots like that already?

almusil:
This is something that should be removed, but this approach has the advantage that we get it right away without the need to wait for new iamge

tinez (approved):

PR #69, opened 2022-03-21
add storage vm to ost runs

hbraha:
Currently, Except for HE, all suites use engine VM for storage,
add a separate VM for storage.

### DIFF
--- a/common/deploy-scripts/setup_storage.sh
+++ b/common/deploy-scripts/setup_storage.sh
@@ -52,8 +52,9 @@ set_selinux_on_nfs() {
 }
 
 setup_iscsi() {
-    # this is ugly, assumes that dedicated storage VMs (ost-[suite]-storage) use their primary network as storage network, and VMs with co-located engine have a dedicated storage network on eth1 (like basic-suite-master).
-    if [[ $(hostname) == *"-storage"* ]]; then
+    # this is ugly, assumes that dedicated storage VMs (ost-[suite]-storage) use their primary network as storage network only on HE suite, and VMs with co-located engine have a dedicated storage network on eth1 (like basic-suite-master).
+    echo $SUITE
+    if [[ $(hostname) == *"-storage"* ]] && [[ $(hostname) == *"he-"*  ]]; then
         NIC=eth0
     else
         NIC=eth1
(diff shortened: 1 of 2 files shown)


common/deploy-scripts/setup_storage.sh
```diff
@@ -52,8 +52,9 @@ set_selinux_on_nfs() {
 }
 
 setup_iscsi() {
-    # this is ugly, assumes that dedicated storage VMs (ost-[suite]-storage) use their primary network as storage network, and VMs with co-located engine have a dedicated storage network on eth1 (like basic-suite-master).
-    if [[ $(hostname) == *"-storage"* ]]; then
+    # this is ugly, assumes that dedicated storage VMs (ost-[suite]-storage) use their primary network as storage network only on HE suite, and VMs with co-located engine have a dedicated storage network on eth1 (like basic-suite-master).
+    echo $SUITE
+    if [[ $(hostname) == *"-storage"* ]] && [[ $(hostname) == *"he-"*  ]]; then
```
tinez, line 57: Maybe it would be better to add the management interface to storage VM in HE suite too?

michalskrivanek: it would be definitely less confusing

common/init-configs/1_engine_2_hosts.json
```diff
@@ -54,6 +56,30 @@
       }
     }
   },
+  "storage": {
```
michalskrivanek, line 59: you should drop these from the engine VM then

michalskrivanek (changes requested):

common/init-configs/1_engine_2_hosts.json
tinez: If we do this in this PR, we'll break OST. Only once all the suites are migrated to use the new storage VM we can do that

common/deploy-scripts/setup_storage.sh
tinez: Let's handle this in subsequent patches

tinez (approved):

michalskrivanek (approved):

PR #73, opened 2022-03-21
use storage vm on network suite

hbraha:
Currently, network suite uses an engine for storage, using the storage VM
added in the previous patch.

### DIFF
--- a/network-suite-master/fixtures/ansible.py
+++ b/network-suite-master/fixtures/ansible.py
@@ -31,6 +31,11 @@ def host1_facts(ansible_host1_facts, af):
     return _machine_facts(ansible_host1_facts.get_all(), af)
 
 
+@pytest.fixture(scope="session")
+def storage_facts(ansible_storage_facts, af):
+    return _machine_facts(ansible_storage_facts.get_all(), af)
+
+
 @pytest.fixture(scope="session", autouse=True)
 def ansible_clean_private_dirs():
     try:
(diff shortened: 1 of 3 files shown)


common/deploy-scripts/setup_storage.sh
```diff
@@ -52,8 +52,9 @@ set_selinux_on_nfs() {
 }
 
 setup_iscsi() {
-    # this is ugly, assumes that dedicated storage VMs (ost-[suite]-storage) use their primary network as storage network, and VMs with co-located engine have a dedicated storage network on eth1 (like basic-suite-master).
-    if [[ $(hostname) == *"-storage"* ]]; then
+    # this is ugly, assumes that dedicated storage VMs (ost-[suite]-storage) use their primary network as storage network only on HE suite, and VMs with co-located engine have a dedicated storage network on eth1 (like basic-suite-master).
+    echo $SUITE
+    if [[ $(hostname) == *"-storage"* ]] && [[ $(hostname) == *"he-"*  ]]; then
```
michalskrivanek, line 57: why is this needed?

hbraha: the HE suite already uses storage vm which have only eth0.

tinez: We're planning to unify the NICes between HE suite and the others once we get these patches in

tinez (approved):

erav (approved):

PR #74, opened 2022-03-21
lagofy: add OST_DEPLOYMENT environment variable

nyoxi:
New environment variable OST_DEPLOYMENT can be used to configure a
different location for deployments. Normally this is in deployment/
subdirectory of the OST repository. But in certain scenarios, e.g. for
storage reasons or to avoid permission issues when repository is on NFS,
it makes sense to provide different location outside of the repository
tree.

Signed-off-by: Tomáš Golembiovský <tgolembi@redhat.com>


tinez (approved):
LGTM, but we definitely need an ack from Michal

michalskrivanek:
we wanted to change the "PREFIX" variable globally. How about doing that now?;)

tr-suite-master/test-run-ocp-installer.sh
```diff
@@ -148,7 +148,7 @@ prefix_for_search="${PREFIX}$(( ( RANDOM % 1000 )  + 1 ))"
 PREFIX=${prefix_for_search:0:14}
 ssh_key=$( cat ${SSH_KEY_FILE} | cut -f1-2 -d " " )
 sed -i  -e 's;SSH_KEY;'"${ssh_key}"';' \
-    -e 's;PREFIX;'"${PREFIX}"';' \
+    -e 's;PREFIX;'"${OST_DEPLOYMENT}"';' \
```
nyoxi, line 151: actually this is wrong

tr-suite-master/test-run-ocp-installer.sh
```diff
@@ -148,7 +148,7 @@ prefix_for_search="${PREFIX}$(( ( RANDOM % 1000 )  + 1 ))"
 PREFIX=${prefix_for_search:0:14}
```
nyoxi, line 148: @emesika: How should this code behave when path in PREFIX is longer than 14 characters? The random suffix will get lost.

michalskrivanek (approved):

tr-suite-master/test-run-ocp-installer.sh
```diff
@@ -13,7 +13,7 @@ Usage: $0 -a ovirt-api-url -u ovirt-user -p ovirt-password -c ovirt-cluster-id -
     -w WORKING_DIR             - working directory for the scripts (def. ${WORKING_DIR})
     -o OPENSHIFT_DIR           - Openshift-install binary directory (def. ${OPENSHIFT_DIR})
     -k SSH_KEY_FILE            - The SSH key file needed by the installer config (def. ${SSH_KEY_FILE})
-    -r PREFIX                  - The prefix that will be used for created master VMs (def. ${PREFIX})
+    -r PREFIX                  - The prefix that will be used for created master VMs (def. ${OST_DEPLOYMENT})
```
tinez, line 16: AFAICT this is [not related](https://github.com/oVirt/ovirt-system-tests/blob/b0411218d3c603418bc4b99b02fd6ab262bc3abe/tr-suite-master/test-scenarios/test_004_run_ocp_installer.py#L86) to OST's prefix. I would omit this file from the patch

nyoxi: That will mean the default is empty string. I suppose that's not a problem as a random number will be appended later. But it would make the help message little bit confusing. Please confirm this is what you intended.

If you are just concerned about resolving the other conversation related to the file we can just ignore that. It was already broken before so nothing changes. Especially in the light of what you said, that it is not a problem for OST run because it provides custom prefix.

michalskrivanek (approved):

PR #76, opened 2022-03-21
basic: Use Selenium grid from storage VM

tinez:
This patch gets rid of all the code that used podman to run Selenium
grid. Instead, we're using the one that exists in 'storage-base' VM.


ost_utils/pytest/fixtures/selenium.py
```diff
@@ -0,0 +1,74 @@
...
+    ip = network_utils.ip_to_url(storage_management_ips[0])
+    url = f"http://{ip}:4444"
+    _grid_health_check(url, 2)
+    yield url
+
+
+@pytest.fixture(scope="session")
+def fetch_videos(ansible_storage, selenium_artifacts_dir, remote_selenium_artifacts_dir):
+    yield
+    ansible_storage.systemd(name="selenium-video-chrome", state="stopped")
```
michalskrivanek, line 67: eh, more ansible, more time and more bugs:/
how about combining the fetch and use wildcard video-*.mp4 ? does that work?
and instead of systemd using shell "systemctl stop both"?

common/init-configs/1_engine_2_hosts.json
```diff
@@ -57,15 +57,15 @@
   },
   "storage": {
     "template": "common/libvirt-templates/vm_template",
-    "memory": "750",
+    "memory": "1792",
```
michalskrivanek, line 60: 1GB more overhead, but it should be all right...

michalskrivanek (approved):

common/init-configs/1_engine_2_hosts.json
tinez: It was used on the executor anyway by the browsers running in containers - we're just shuffling the reqs around.

ost_utils/pytest/fixtures/selenium.py
tinez: Unfortunately `fetch` doesn't support wildcards. I will combine the two `stop` calls into a single `shell` invocation though.

ljelinkova (commented):
Looks good, just one question.

basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
```diff
@@ -180,7 +184,7 @@ def browser_name(browser_options):
 
 
 @pytest.fixture(scope="session")
-def ovirt_driver(browser_options, hub_url, engine_webadmin_url):
+def ovirt_driver(browser_options, hub_url, fetch_videos, engine_webadmin_url, selenium_artifacts_dir):
     driver = webdriver.Remote(command_executor=hub_url, options=browser_options)
```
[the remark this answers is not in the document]
tinez: Hmm looks like a leftover from previous patch version, good catch, thanks!

ljelinkova (approved):

hbraha (approved):

PR #77, opened 2022-03-23
github: Fix custom repo artifact collection

acernek:
Some projects have multiple runs, not all of which include artifacts.
We should therefore check for artifacts during the search for workflow
run corresponding to specific commit.

### DIFF
--- a/ost_utils/deployment_utils/package_mgmt.py
+++ b/ost_utils/deployment_utils/package_mgmt.py
@@ -153,7 +153,13 @@ def _github_resolve_commit_to_workflow_run(repo, commit) -> str:
     )
     for run in runs_response.json()["workflow_runs"]:
         if run["head_sha"] == commit:
-            return run["id"]
+            run_id = run["id"]
+            artifacts_response = _github_get(
+                f"https://api.github.com/repos/oVirt/{repo}/actions/runs/{run_id}/artifacts"
+            ).json()
+            for artifact in artifacts_response["artifacts"]:
+                if not artifact["expired"] and artifact["name"].startswith("rpm"):
+                    return run_id
 
     raise RuntimeError(f"No workflow runs found for commit {commit}")
 


michalskrivanek:
any PR to test this with?

michalskrivanek:
LGTM but not sure about verification

tinez:
I've picked https://github.com/oVirt/vdsm/pull/111 as a candidate - just triggered a [code quality workflow](https://github.com/oVirt/vdsm/actions/runs/2061731793) on it. When it's finished we should try triggering OST on it - I hope it will fail. Then we can verify the same with this PR.

ost_utils/deployment_utils/package_mgmt.py
```diff
@@ -153,7 +153,12 @@ def _github_resolve_commit_to_workflow_run(repo, commit) -> str:
     )
     for run in runs_response.json()["workflow_runs"]:
         if run["head_sha"] == commit:
-            return run["id"]
+            run_id = run["id"]
+            artifacts_response = _github_get(
+                f"https://api.github.com/repos/oVirt/{repo}/actions/runs/{run_id}/artifacts"
+            ).json()
+            if artifacts_response["artifacts"]:
```
mwperina, line 160: Shouldn't we check the existence of `rpm-<PLATFORM>` artifacts as created within https://github.com/oVirt/upload-rpms-action ?

tinez: > Shouldn't we check the existence of `rpm-<PLATFORM>` artifacts as created within https://github.com/oVirt/upload-rpms-action ?

Yeah, if we want a good solution we would need to gather `run_id`s and then check for presence of artifacts, their expiration and pattern match for distro names. As I said - this patch is good as a quick fix (under the assumption that we don't have projects with non-RPM workflows that produce artifacts).

acernek: I've verified the patch locally, but I also can try to improve on it, whichever suits you.

acernek: Done

acernek: Marcin would also like to avoid duplicate http requests, so I've made an alternative patch: https://github.com/acernek/ovirt-system-tests/commit/3f73b104bcba38fdafc84f41243b9401c35c3159 Let me know if it's better.

michalskrivanek: ok, let's leave it for later patch

PR #79, opened 2022-03-24
Adding go-ovirt-client-tests suite

emesika:
This patch adds a ansible based suite named
go-ovirt-client-tests-suite-master that will test oVirt go API via the
go-ovirt.

Signed-off-by: Eli Mesika <emesika@redhat.com>


michalskrivanek (changes requested):
I don't particularly like scripts that just runs a script. just define the env vars and run go-ovirt-client-tests-exe directly

also, IIUC it needs to be installed somehow?

also, what about the ocp installer test?

janosdebugs:
@michalskrivanek the go-ovirt-client test is a single binary that just needs to be run with the [proper environment variables set](https://github.com/oVirt/go-ovirt-client/blob/main/CONTRIBUTING.md#running-tests). We do not need the installer integration to be tested, we just need our own test suite to be run. (It is already way more extensive than the installer could ever be.)

michalskrivanek:
i think it's ok to identify which case to run based on what's available.  go-ovirt-client would be available only if you run with those artifacts - in that case it would make sense to skip the current ocp installer code and run just this one.
unless you say we should abandon the installer test altogether

janosdebugs:
The installer tests will be useful in the future, but only once we have the new version integrated. Until then, we will have breaking changes so that test will fail anyway.

michalskrivanek:
@emesika hi eli, any update?

.github/jenkins
```diff
@@ -12,7 +12,7 @@ PR_URL="$(sed -n 1p queue)"
 SUITE="$(sed -n 2p queue)"
 DISTRO="$(sed -n 3p queue)"
 echo "$PR_URL" | egrep '^https://(api.|www.)?github.com(/repos)?/oVirt/'
-echo "$SUITE"  | egrep '^(basic-suite-master|he-basic-suite-master|ansible-suite-master|tr-suite-master|network-suite-master|tr-suite-master)$'
+echo "$SUITE"  | egrep '^(basic-suite-master|he-basic-suite-master|ansible-suite-master|tr-suite-master|network-suite-master|go-ovirt-client-tests-suite-master)$'
```
michalskrivanek, line 15: it's a terrible name. "go-suite-master" would be better.
Also if it's replacing t suite then drop it from everywhere please

common/deploy-scripts/setup_engine.sh
```diff
@@ -34,5 +34,5 @@ cp /usr/share/doc/ovirt-engine/mibs/* /usr/share/snmp/mibs
 systemctl start snmptrapd
 systemctl enable snmptrapd
 
-# TODO workaround for https://bugzilla.redhat.com/show_bug.cgi?id=2077794
```
michalskrivanek, line 37: don't remove this

michalskrivanek (changes requested):
for tracking history it would be better to rename stuff from tr-suite-master and not create new copies

.github/jenkins
emesika: Done

common/deploy-scripts/setup_engine.sh
emesika: Done

go-suite-master/test-scenarios/test_003_run_go_ovirt_client_tests.py
```diff
@@ -0,0 +1,27 @@
...
+    output_file = os.path.join(artifacts_dir, 'go-ovirt-client-tests.out')
+    with open(output_file, 'w') as o:
+        lines = ansible_engine.shell(
+            f"OVIRT_CAFILE=/etc/pki/ovirt-engine/ca.pem \
+              OVIRT_URL={engine_api_url} \
+              OVIRT_USERNAME={engine_full_username} \
+              OVIRT_PASSWRD={engine_password} \
+              go-ovirt-client-tests-exe"
+        )['stdout_lines']
+        for line in lines:
```
tinez, line 26: You don't even need to iterate here - just use `stdout` instead of `stdout_lines` and pass it directly to `o.write`

emesika: Done

go-suite-master/test-scenarios/test_003_run_go_ovirt_client_tests.py
```diff
@@ -0,0 +1,26 @@
...
+                           engine_full_username, engine_password, artifacts_dir):
+
+    ansible_engine.shell("dnf install -y go-ovirt-client-tests")
+    output_file = os.path.join(artifacts_dir, 'go-ovirt-client-tests.out')
+    with open(output_file, 'w') as o:
+        out = ansible_engine.shell(
+            f"OVIRT_CAFILE=/etc/pki/ovirt-engine/ca.pem \
+              OVIRT_URL={engine_api_url} \
+              OVIRT_USERNAME={engine_full_username} \
+              OVIRT_PASSWRD={engine_password} \
```
michalskrivanek, line 23: OVIRT_PASSWORD

common/deploy-scripts/setup_engine.sh
```diff
@@ -34,3 +34,5 @@ cp /usr/share/doc/ovirt-engine/mibs/* /usr/share/snmp/mibs
 systemctl start snmptrapd
 systemctl enable snmptrapd
 
+# workaround for https://bugzilla.redhat.com/show_bug.cgi?id=2077794
+rpm -q rhvm || dnf downgrade -y --repo appstream postgresql-jdbc-42.2.3
```
michalskrivanek, line 38: please remove/rebase

common/deploy-scripts/setup_engine.sh
```diff
@@ -33,4 +33,3 @@ cp /usr/share/doc/ovirt-engine/mibs/* /usr/share/snmp/mibs
 
 systemctl start snmptrapd
 systemctl enable snmptrapd
-
```
michalskrivanek, line 36: you don't need to touch this file at all;)

tinez (approved):

michalskrivanek (approved):

PR #85, opened 2022-03-30
Use nobest in 'DNF update the system'

didib:
Without this, it fails, because it wants to update
ovirt-ansible-collection, which now requires
ansible-collection-ansible-utils, which requires ansible-core, whereas
the engine still conflicts with ansible-core.

TODO: Revert this (and perhaps some more) when the engine is fixed to
work with ansible-core.

Change-Id: I4851b76a353edd274cb4cfc76a0467b117f443c4

### DIFF
--- a/common/deploy-scripts/setup_first_he_host.sh
+++ b/common/deploy-scripts/setup_first_he_host.sh
@@ -72,6 +72,7 @@ dnf_update() {
     name:  "*"
     state: latest
     enablerepo: "rhel*-baseos,rhel*-appstream"
+    nobest: yes
     exclude:
       - ovirt-release-master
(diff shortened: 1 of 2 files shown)


didib:
@mnecas @mwperina @dangel101 FYI. With this, we skip updating ovirt-ansible-collection in he-basic. I hope your recent changes there are more for hosts than for the engine. If not, we should not merge this but try to quickly fix the engine. I see we have a few PRs there, not sure about status.

michalskrivanek:
would you please post a revert?

michalskrivanek:
or there's a button for that!:)

didib:
> or there's a button for that!:)

At least this github does have...

PR #88, opened 2022-03-30
ansible: Change arg order passed to ansible-runner

tinez:
New ansible-runner versions require '-p' argument to be passed after the
'run' command.


mnecas (approved):
lgtm

PR #90, opened 2022-03-31
github: Improve custom repo expansion

acernek:
Avoid duplicated HTTP requests, while checking multiple workflow runs.

### DIFF
--- a/ost_utils/deployment_utils/package_mgmt.py
+++ b/ost_utils/deployment_utils/package_mgmt.py
@@ -60,28 +60,44 @@ def expand_github_repo(repo_url, working_dir, ost_images_distro):
     if not match:
         raise RuntimeError(f"Not a valid GitHub link {repo_url}")
     _, _, repo, _, pr, commit, workflow_run, _ = match.groups()
-    if not workflow_run:
+
+    LOGGER.debug("GitHub link: repo=%s pr=%s commit=%s workflow_run=%s", repo, pr, commit, workflow_run)
+
+    if workflow_run:
+        workflow_runs = [workflow_run]
+    else:
         if not commit:
             commit = _github_resolve_pr_to_commit(repo, pr)
-        workflow_run = _github_resolve_commit_to_workflow_run(repo, commit)
+            LOGGER.debug("Commit %s in %s repo found for pr %s", commit, repo, pr)
+        workflow_runs = _github_resolve_commit_to_workflow_runs(repo, commit)
+        LOGGER.debug("Number of workflow runs found for commit %s: %d", commit, len(workflow_runs))
 
-    artifacts: list[_GitHubArtifact] = _github_list_artifacts(repo, workflow_run)
-    for artifact in artifacts:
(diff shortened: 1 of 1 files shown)


ost_utils/deployment_utils/package_mgmt.py
```diff
@@ -60,28 +60,40 @@ def expand_github_repo(repo_url, working_dir, ost_images_distro):
...
+    for run_id in workflow_runs:
+        artifacts: list[_GitHubArtifact] = _github_list_artifacts(repo, run_id)
+        LOGGER.debug("Workflow run %s: %d artifacts", run_id, len(artifacts))
+        for artifact in artifacts:
+            LOGGER.debug("Artifact: %s", artifact.name)
+            if "rpm" in artifact.name and OST_TO_GITHUB_DISTRO_NAME[ost_images_distro] in artifact.name:
+                target_path = os.path.join(working_dir, 'github_artifacts', f'{commit}-{artifact.name}')
+                os.makedirs(target_path, exist_ok=True)
+                # Download the artifact
+                target_file = _github_download_artifact(artifact, target_path)
```
tinez, line 84: Let's have like 3 more debugging messages for starting the download, unpacking and the "has_rpm" check

tinez (commented):
Looks good, see my small suggestion

ost_utils/deployment_utils/package_mgmt.py
```diff
@@ -60,28 +60,44 @@ def expand_github_repo(repo_url, working_dir, ost_images_distro):
...
-                return target_path
-    if not artifacts:
-        raise RuntimeError(f"GH pr/commit/run {repo_url} had no artifacts with RPM files.")
-    raise RuntimeError(f"GH pr/commit/run {repo_url} didn't match any of the artifacts names to distro name.")
+            LOGGER.debug("Commit %s in %s repo found for pr %s", commit, repo, pr)
+        workflow_runs = _github_resolve_commit_to_workflow_runs(repo, commit)
+        LOGGER.debug("Number of workflow runs found for commit %s: %d", commit, len(workflow_runs))
+
+    for run_id in workflow_runs:
+        artifacts: list[_GitHubArtifact] = _github_list_artifacts(repo, run_id)
```
acernek, line 76: BTW, this will raise an exception in the case of expired artifacts - can older runs have working artifacts or is there an assurance that all older runs are bound to have expired artifacts as well?

tinez: AFAIR older runs can have working artifacts - the artifact expiration time is a separate setting and it's not strictly related to the lifetime of the run itself.

acernek: So should I remove the exception? (or rather replace it with logging?)

tinez: Let's replace it with logging

acernek:
Verified on VDSM patch with runs without artifacts

PR #102, opened 2022-04-06
Engine's keycloak internal authentication support

arso:
For upstream (el8stream) use engine credentials provided by internally
bundled Keycloak instance.

For downstream tests (rhel) stick to the legacy AAA

### DIFF
--- a/network-suite-master/fixtures/providers.py
+++ b/network-suite-master/fixtures/providers.py
@@ -15,7 +15,6 @@ from testlib import suite
 
 
 OPENSTACK_AUTH_URL = 'https://{}:35357/v2.0'
-OPENSTACK_USERNAME = 'admin@internal'
 OVIRT_IMAGE_REPO_NAME = 'ovirt-image-repository'
 OVIRT_IMAGE_REPO_URL = 'http://glance.ovirt.org:9292/'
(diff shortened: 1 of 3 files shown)


network-suite-master/fixtures/providers.py
```diff
@@ -15,7 +15,7 @@
 
 
 OPENSTACK_AUTH_URL = 'https://{}:35357/v2.0'
-OPENSTACK_USERNAME = 'admin@internal'
+OPENSTACK_USERNAME = 'admin@ovirt@internalsso'
```
erav, line 18: is this not dependent on the distro?

arso: yes it is !  thanks Eitan. I will  add the conditions to support it.

erav (approved):

arso:
OST run without a fix for Eitan's comment was successful https://rhv-devops-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/ds-ost-baremetal_manual/33110/console
The mentioned fix is only for rhel so I think that this all that had to be done here for networking suite.

erav:
host-1 install failed. setup networks could net set up ovirtmgmt network. only thing i can find in the failed run logs is MOM not available on host-1 vdsm.log. don't know if its related. rerunning

PR #104, opened 2022-04-07
Fix keycloak_enabled logic

didib:
Without this patch, it fails on hosted-engine, where we currently
disabled it. Change the logic to check inside the engine machine instead
of guessing.

Change-Id: I681038351471af4747f61ed017fbf8c8e9a8d52a


michalskrivanek:
bleh, another ansible...but yeah, makes sense we rather read the actual state

PR #108, opened 2022-04-07
he: another workaround for updating RHV appliance

michalskrivanek:
ansible's dnf module has an interesting behavior. Good old shell
actually works.


michalskrivanek:
rhv works

PR #109, opened 2022-04-08
Fix keycloak_enabled to work better

michalskrivanek:
otopi-config-query is great, but sometimes we have a single
config, sometimes we use conf.d subdirectory. KISS and just grep.


mwperina (approved):
Let's try

PR #115, opened 2022-04-12
basic: remove async_ argument when calling ExternalVmImportService

nyoxi:
There is no `async_` argument for the `add()` call in `ExternalVmImportService`. The argument is silently ignored and having that in the call is misleading/confusing to anyone reading the code.

Signed-off-by: Tomáš Golembiovský <tgolembi@redhat.com>


nyoxi:
I am not entirely convinced we want to merge this. The PR may server us as a reminder to fix it in the API. What do you think @ahadas?

tinez (approved):

michalskrivanek:
...better to merge than keeping a stale reminder:)

PR #118, opened 2022-04-13
he: add storage network for vms

hbraha:
To maintain uniformity across all suites using storage VMs, add storage network and NICs.

### DIFF
--- a/common/deploy-scripts/setup_storage.sh
+++ b/common/deploy-scripts/setup_storage.sh
@@ -52,13 +52,7 @@ set_selinux_on_nfs() {
 }
 
 setup_iscsi() {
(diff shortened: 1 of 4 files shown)


didib (approved):
I do not fully know how lagofy networks/nics works, but patch looks ok. Please verify also with ISCSI HE storage - but can be merged before that, because we do not (yet?) run this routinely.

tinez:
> I do not fully know how lagofy networks/nics works, but patch looks ok. Please verify also with ISCSI HE storage - but can be merged before that, because we do not (yet?) run this routinely.

@hbraha could you please try to run HE with iscsi too?

PR #128, opened 2022-04-25
setup: replace ansible modules with shell

michalskrivanek:
ansible 2.10+ moved around the roles, it's incomatible with ansible 2.9
that we still support and/or requires community module. Easiest is to
just call shel


common/setup/setup_playbook.yml
```diff
@@ -162,17 +162,11 @@
...
+      shell: semanage permissive -a virtlogd_t
       become: true
 
     - name: allow port 2222 for IPv6 socks proxy
-      ansible.posix.firewalld:
-        zone: libvirt
-        port: 2222/tcp
-        permanent: yes
-        state: enabled
+      shell: firewall-cmd --add-port 2222/tcp --zone=libvirt --permanent
```
mwperina, line 169: We have included ansible-collection-ansible-posix package, which includeds firewalld module, into both oVirt/RHV, so no need to replace, just fix requirements

mwperina (changes requested):

common/setup/setup_playbook.yml
michalskrivanek: sure, but i'm done with ansible. the way it breaks all the time they change something, and the next-to-impossible debugging.
This is not running on oVirt/RHV machine but on "bare" el[89]stream/RHEL 8.6, dealing with additional special" repos an extra hassle

PR #132, opened 2022-04-25
lagofy: Fix ansible-galaxy setup

tinez:
8.6 uses 'ansible-core' and requires the additional galaxy packages
to be installed too.

### DIFF
--- a/lagofy.sh
+++ b/lagofy.sh
@@ -340,7 +340,7 @@ ost_check_dependencies() {
         namei -vm `pwd`
         return 8
     }
-    (source /etc/os-release; [[ $VERSION == 9* ]]) && { ansible-galaxy collection install community.general openstack.cloud >/dev/null || {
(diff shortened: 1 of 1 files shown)


michalskrivanek (approved):
verified

lagofy.sh
```diff
@@ -340,7 +340,7 @@ ost_check_dependencies() {
         namei -vm `pwd`
         return 8
     }
-    (source /etc/os-release; [[ $VERSION == 9* ]]) && { ansible-galaxy collection install community.general openstack.cloud >/dev/null || {
+    (rpm -q ansible-core &> /dev/null) && { ansible-galaxy collection install community.general openstack.cloud >/dev/null || {
```
michalskrivanek, line 343: actually, also drop the subshell (), not needed

tinez: done

PR #133, opened 2022-04-26
Test, do not merge: Enable keycloak

didib:


ost_utils/pytest/fixtures/engine.py
```diff
@@ -76,7 +76,7 @@ def engine_username(keycloak_enabled):
 @pytest.fixture(scope="session")
 def engine_full_username(keycloak_enabled):
     if keycloak_enabled:
-        return "admin@ovirt@internalsso"
```
didib, line 79: @arso please see this. A previous run failed, so I pushed this patch. Also ran a test with it, still didn't finish, I am going to sleep. But I thought: This does not make sense. Why? Because in the patch to HE deploy code, I did use admin@ovirt@internalsso [1], and it does work there (when it failed, it failed later then deploy). So perhaps the issue is elsewhere? E.g.: HE deploy code is fully ansible, using the ovirt_auth module (itself obviously written in python). But OST is not - at least the engine_api fixture is using the python sdk directly. So perhaps there is a difference around this?

[1] https://github.com/oVirt/ovirt-ansible-collection/pull/488/commits/869242a7ccc61d6cf4faa80473b8bced668c5c89

didib: Well, it failed also with this patch, with admin@ovirt. So it's something else. I'll drop this patch (or perhaps revert, for now, to keep my comments). Runs:

With current patch: https://redir.apps.ovirt.org/dj/job/ds-ost-baremetal_manual/35168/

Without it: https://redir.apps.ovirt.org/dj/job/ds-ost-baremetal_manual/35158/

didib: OK, please ignore. I had a bug in my OST patch for installing keycloak on the engine VM, and it was not installed. So deploy code worked correctly with admin@internal.

arso:
Deprecated by https://github.com/oVirt/ovirt-system-tests/pull/141

didib:
Great job!

PR #136, opened 2022-05-01
run vms on el9

hbraha:
We need to create the VMs without QXL devices in order to run them on EL9.
The patch removes devices that use spice/qxl.

### DIFF
--- a/basic-suite-master/test-scenarios/test_002_bootstrap.py
+++ b/basic-suite-master/test-scenarios/test_002_bootstrap.py
@@ -1218,13 +1218,13 @@ def test_add_blank_vms(engine_api, ost_cluster_name):
             name=TEMPLATE_BLANK,
         ),
         display=sdk4.types.Display(
-            smartcard_enabled=True,
             keyboard_layout='en-us',
             file_transfer_enabled=True,
             copy_paste_enabled=True,
+            type=sdk4.types.DisplayType.VNC,
         ),
         usb=sdk4.types.Usb(
-            enabled=True,
+            enabled=False,
             type=sdk4.types.UsbType.NATIVE,
         ),
         memory_policy=sdk4.types.MemoryPolicy(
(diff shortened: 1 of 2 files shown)


basic-suite-master/test-scenarios/test_002_bootstrap.py
```diff
@@ -1218,13 +1218,14 @@ def test_add_blank_vms(engine_api, ost_cluster_name):
...
         display=sdk4.types.Display(
-            smartcard_enabled=True,
+            smartcard_enabled=False,
             keyboard_layout='en-us',
             file_transfer_enabled=True,
             copy_paste_enabled=True,
+            type=sdk4.types.DisplayType.VNC,
         ),
         usb=sdk4.types.Usb(
-            enabled=True,
```
michalskrivanek, line 1227: this will disable smartcard and usb for all platforms, but that's ok, not so interesting

basic-suite-master/test-scenarios/test_002_bootstrap.py
```diff
@@ -1444,23 +1445,25 @@ def test_add_nic(engine_api):
 
 
 @order_by(_TEST_LIST)
-def test_add_graphics_console(engine_api):
+def test_add_graphics_console(engine_api, ansible_host0_facts):
+    if ansible_host0_facts.get('ansible_distribution_major_version') == "9":
+        pytest.skip('skip test on el9')
```
michalskrivanek, line 1450: why? don't you want to use VNC?

basic-suite-master/test-scenarios/test_002_bootstrap.py
```diff
@@ -1444,23 +1445,25 @@ def test_add_nic(engine_api):
...
     # remove VNC
     engine = engine_api.system_service()
     vm = test_utils.get_vm_service(engine, VM0_NAME)
     consoles_service = vm.graphics_consoles_service()
-    if len(consoles_service.list()) == 2:
+    consoles = consoles_service.list()
+    console = next((c for c in consoles if c.protocol == types.GraphicsType.VNC), None)
+    if console is not None:
```
michalskrivanek, line 1457: the point here would be to test there *is* a VNC console, not a code that handles various states.

basic-suite-master/test-scenarios/test_004_basic_sanity.py
```diff
@@ -635,7 +635,9 @@ def test_preview_snapshot_with_memory(engine_api):
 
 
 @order_by(_TEST_LIST)
-def test_vmconsole(engine_api, engine_ip, working_dir, rsa_pair):
+def test_vmconsole(engine_api, engine_ip, working_dir, rsa_pair, ansible_host0_facts):
+    if ansible_host0_facts.get('ansible_distribution_major_version') == "9":
+        pytest.skip('el9 dont use vmconsole')
```
michalskrivanek, line 640: it should;)

basic-suite-master/test-scenarios/test_004_basic_sanity.py
```diff
@@ -938,6 +940,19 @@ def test_cold_incremental_backup_vm2(engine_api, get_vm_service_for_vm):
     backup.perform_incremental_vm_backup(engine_api, vm2_backups_service, DISK2_NAME, "cold_vm_backup")
 
 
+def add_selinux_allow_rule_for_ssh(ansible_hosts, root_dir):
```
michalskrivanek, line 943: why would we add anything like that? what is it trying to fix?

michalskrivanek (changes requested):

basic-suite-master/test-scenarios/test_002_bootstrap.py
hbraha: after deleting and re-adding the VNC the video type chosen automatically is QXL which is not supported on el9. this should be resolved as part of dropping the QXL in the future.

basic-suite-master/test-scenarios/test_002_bootstrap.py
hbraha: sorry, I think dont really understand what you meant here.
 the code here as the previous one, checking if vnc is there, removing it, and re-adding it. what exactly is the difference in the behviour?

basic-suite-master/test-scenarios/test_004_basic_sanity.py
hbraha: ack, fixed

basic-suite-master/test-scenarios/test_004_basic_sanity.py
hbraha: not needed anymore, the fix should be shipped with the new vmconsole package that was posted this morning.

basic-suite-master/test-scenarios/test_002_bootstrap.py
michalskrivanek: hm, i don't really know what i meant, nevermind:)

michalskrivanek (approved):

michalskrivanek:
actually, please improve the commit message. It's not very...helpful

PR #140, opened 2022-05-04
ost_utils: vmconsole - use default byte stream

erav:
Align with PR#135 [1]:

By default 'Popen' processes's stdout is opened as a byte stream.
Passing the 'universal_newlines=True' argument makes the stream
automatically decoded from UTF-8.

Since we're reading the output char by char, multibyte UTF-8 characters
can break the stream with i.e.:

E UnicodeDecodeError: 'utf-8' codec can't decode byte 0xda ...
invalid continuation byte

Let's change the output to have a form of bytes and use bytes in
comparisons too.

[1] https://github.com/oVirt/ovirt-system-tests/pull/135

Change-Id: I799faaa6617b7a1f48ce9175509eef1d7e62c4de


erav:
@arso : I see errors in engine.log:
` OAuthException access_denied: Cannot authenticate user Invalid user credentials `
and in keycloak.log:
` error=user_not_found, auth_method=openid-connect, grant_type=password, client_auth_method=client-secret, username=admin `
Can you advise? Thanks

arso:
For some reason  legacy admin@internal is being used in 'test_bond_active_slave'  test. Investigating why. 
https://rhv-devops-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/ds-ost-baremetal_manual/36332/console#L150

PR #142, opened 2022-05-06
selenium: Use standalone nodes running one at a time

tinez:
In an attempt to limit the resources needed to run Selenium we're now using
the "standalone" images and running a single browser at a time.

[Related change](https://github.com/oVirt/ost-images/pull/53) on ost-images side.

Change-Id: Ic57391f70e22fb07bbbf441e7874fbf406d71232


michalskrivanek (approved):
Considering it's now broken almost always, no need to wait. This will mean mo video for now, right? Still better passing without video than failing...

PR #151, opened 2022-05-16
ost_setup: remove Ansible installation

hbraha:
since we use the ansible-core package in the playbook setup, Installing
Ansible can cause a dependency issue.

### DIFF
--- a/setup_for_ost.sh
+++ b/setup_for_ost.sh
@@ -51,17 +51,6 @@ fi
 source /etc/os-release
 
 ANSIBLE_INSTALLED=$(which ansible-playbook &> /dev/null && echo 1 || echo 0)
-ANSIBLE_PACKAGE=$(if [[ $VERSION == 9* ]]; then echo "ansible-core"; else echo "ansible"; fi)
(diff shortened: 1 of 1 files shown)


setup_for_ost.sh
```diff
@@ -51,17 +51,7 @@ fi
...
-ANSIBLE_REPO_RPM_NAME=${ANSIBLE_REPO_RPM_NAME:-"epel-release-8"}
-
-cleanup() {
-    if [[ ${REMOVE_ANSIBLE_REPO_RPM} -eq 1 ]]; then
-        sudo dnf remove -y "${ANSIBLE_REPO_RPM_NAME}"
-    fi
-}
-
-trap cleanup EXIT
+ANSIBLE_PACKAGE=$(echo "ansible-core")
```
michalskrivanek, line 54: so we dont' really need ANSIBLE_PACKAGE anymore, right

michalskrivanek (changes requested):

setup_for_ost.sh
hbraha: yes indeed, this is just in order to have the package name in a variable, would you prefer it to be hardcoded?

michalskrivanek: yes, I think it's very simple code, no need for variable

PR #152, opened 2022-05-16
coverage: Add script for merging from recent downstream run

acernek:


.coveragerc
```diff
@@ -0,0 +1,10 @@
+[paths]
+ost = 
+    ./
+    /home/jenkins/workspace/ds-ost-baremetal_manual/
+    /home/jenkins/workspace/ds-ost-baremetal_manual@1/
```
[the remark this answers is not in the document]
acernek: Marcin wants to create it on-the-fly on jenkins side, this is just for testing

common/scripts/coverage_merge.sh
```diff
@@ -0,0 +1,77 @@
...
+get_dist() {
+    echo $(echo $1 | grep -oP "OST_IMAGE=\K\w+(?=\+)")
+}
+
+get_coverage_name() {
+    IFS='/' read -r -a split_coverage <<< "$1"
+    echo "${split_coverage[2]}"
+}
+
+[[ $# -eq 0 ]] && { echo "URL to downstream baremetal OST run expected"; exit 1; }
```
tinez, line 29: I think that we should say we need an URL to a `ds-ost-baremetal-ost` job explicitly - with the current message I'd think we mean a `ds-ost-baremetal_manual` job

common/scripts/coverage_merge.sh
```diff
@@ -0,0 +1,77 @@
...
+    coverage_name=$(get_coverage_name $run_coverage)
+    
+    echo "Downloading the coverage file $coverage_name..."
+    wget -q -nc -O "$merge_dir/$run_distro.$coverage_name" "${run_url}/artifact/${run_coverage}"
+
+    echo "---"
+
+done
+
+echo "Preparing for the coverage merge..."
```
tinez, line 71: We also need to `git checkout ${merged_git}` here somewhere before we start creating the report.

acernek: Done, but maybe will need some suppression of git output? (they don't seem too distracting to me)

tinez: I think it's fine even with git output

PR #157, opened 2022-05-17
coverage: Set vdsm coverage using pytest option

acernek:


tinez (approved):
Looks good. We need to sync that with jenkins pipelines. I'll handle this.

tinez (approved):

PR #158, opened 2022-05-17
run el9 hosts with el8 engine

hbraha:
As we currently don't have an el9 engine image, this hack will provide the
option to run el9 hosts with el8 engine.

### DIFF
--- a/common/helpers/ost-images.sh
+++ b/common/helpers/ost-images.sh
@@ -23,6 +23,11 @@ OST_IMAGES_HOST_INSTALLED=${OST_IMAGES_HOST_INSTALLED:-${OST_IMAGES_DIR}/${OST_I
 OST_IMAGES_HE_INSTALLED=${OST_IMAGES_HE_INSTALLED:-${OST_IMAGES_DIR}/${OST_IMAGES_DISTRO}-he-installed.qcow2}
 OST_IMAGES_STORAGE_BASE=${OST_IMAGES_STORAGE_BASE:=${OST_IMAGES_DIR}/storage-base.qcow2}
 OST_IMAGES_SSH_KEY=${OST_IMAGES_SSH_KEY:-${OST_IMAGES_DIR}/${OST_IMAGES_DISTRO}_id_rsa}
+# FIXME: until we will build el9stream enigne image, we can use el8 engine with el9 hosts
+if [[ "$OST_IMAGES_DISTRO" == "el9stream" ]]; then
+  OST_IMAGES_ENGINE_INSTALLED=${OST_IMAGES_ENGINE_INSTALLED/el9stream/el8stream}
+  OST_IMAGES_NODE=${OST_IMAGES_HOST_INSTALLED:-${OST_IMAGES_DIR}/${OST_IMAGES_DISTRO}-host-installed.qcow2}
+fi
 for i in OST_IMAGES_BASE OST_IMAGES_NODE OST_IMAGES_ENGINE_INSTALLED OST_IMAGES_HOST_INSTALLED OST_IMAGES_HE_INSTALLED OST_IMAGES_SSH_KEY; do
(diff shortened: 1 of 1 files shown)


common/helpers/ost-images.sh
```diff
@@ -23,6 +23,10 @@ OST_IMAGES_HOST_INSTALLED=${OST_IMAGES_HOST_INSTALLED:-${OST_IMAGES_DIR}/${OST_I
 OST_IMAGES_HE_INSTALLED=${OST_IMAGES_HE_INSTALLED:-${OST_IMAGES_DIR}/${OST_IMAGES_DISTRO}-he-installed.qcow2}
 OST_IMAGES_STORAGE_BASE=${OST_IMAGES_STORAGE_BASE:=${OST_IMAGES_DIR}/storage-base.qcow2}
 OST_IMAGES_SSH_KEY=${OST_IMAGES_SSH_KEY:-${OST_IMAGES_DIR}/${OST_IMAGES_DISTRO}_id_rsa}
+# FIXME: until we will build el9stream enigne image, we can use el8 engine with el9 hosts
+if [[ "$OST_IMAGES_ENGINE_INSTALLED" == *"el9stream"* ]]; then
+  OST_IMAGES_ENGINE_INSTALLED=${OST_IMAGES_ENGINE_INSTALLED/el9stream/el8stream}
```
tinez, line 28: I think we also need to change `OST_IMAGES_NODE="${OST_IMAGES_HOST_INSTALLED}"` in this case.

michalskrivanek (changes requested):
I don't think this file is the best place. It allows for setting the variable up front as an override. And now you'd break that and replace user-supplied variable.

I think the change should be done in the caller - ost.sh or even just the jenkins job definition.

common/helpers/ost-images.sh
```diff
@@ -23,6 +23,11 @@ OST_IMAGES_HOST_INSTALLED=${OST_IMAGES_HOST_INSTALLED:-${OST_IMAGES_DIR}/${OST_I
 OST_IMAGES_HE_INSTALLED=${OST_IMAGES_HE_INSTALLED:-${OST_IMAGES_DIR}/${OST_IMAGES_DISTRO}-he-installed.qcow2}
 OST_IMAGES_STORAGE_BASE=${OST_IMAGES_STORAGE_BASE:=${OST_IMAGES_DIR}/storage-base.qcow2}
 OST_IMAGES_SSH_KEY=${OST_IMAGES_SSH_KEY:-${OST_IMAGES_DIR}/${OST_IMAGES_DISTRO}_id_rsa}
+# FIXME: until we will build el9stream enigne image, we can use el8 engine with el9 hosts
+if [[ "$OST_IMAGES_ENGINE_INSTALLED" == *"el9stream"* ]]; then
```
tinez, line 27: Maybe using `OST_IMAGES_DISTRO` here would make more sense right now since we're modifying vars for more than one image.

michalskrivanek:
> > I don't think this file is the best place. It allows for setting the variable up front as an override. And now you'd break that and replace user-supplied variable.
> 
> There's no other way of testing el9stream right now than the one we enforce here - we have neither an el9stream engine image nor an el9stream-based node image available.

we do have el9stream based node, adding ost image is trivial. we just haven't done this since basic site doesn't pass yet

> This is a temporary hack for convenience. Once we have the other images available we'll get rid of that.
> 
> > I think the change should be done in the caller - ost.sh or even just the jenkins job definition.
> 
> If we put it in jenkins people will still have to modify OST code for local runs, which is annoying. If we'd like to put in `ost.sh` we'd need to replicate some logic from `ost-images.sh`. I vote for keeping it here.

it's annoying if you take my override and change it into something i do not want to run, without an option not to do that.
i.e. do OST_IMAGES_ENGINE_INSTALLED=${OST_IMAGES_ENGINE_INSTALLED:-${OST_IMAGES_DIR}/${OST_IMAGES_DISTRO/el9stream/el8stream}-engine-installed.qcow2} instead

PR #159, opened 2022-05-18
ost: downgrade nmstate

hbraha:
workaround for BZ2084118, downgrading nmstate at el9 hosts.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2084118

### DIFF
--- a/common/deploy-scripts/setup_host.sh
+++ b/common/deploy-scripts/setup_host.sh
@@ -22,6 +22,16 @@ if [[ $(which nodectl) ]]; then
     echo 3 > /proc/sys/vm/drop_caches
 fi
 
+# FIXME: work around for BZ 2084118 - Downgrading nmstate
+DISTRO=$( cat /etc/redhat-release )
+if [[ "$DISTRO" == *"9"* ]]; then
+    rpm -e --nodeps nmstate
+    dnf config-manager --set-enabled appstream
+    dnf install -y nmstate-2.0.0-0.4.alpha3.el9
+    echo exclude="nmstate nmstate-plugin-ovsdb python3-libnmstate" >> /etc/dnf/dnf.conf
+    dnf config-manager --set-disabled appstream
+fi
+
 # Set up hugepages
 HUGEPAGES=3
 for node in /sys/devices/system/node/node*; do


ost_utils/pytest/fixtures/deployment.py
```diff
@@ -118,6 +119,14 @@ def deploy(
     # set static hostname to match the one assigned by DNS
     ansible_all.shell("hostnamectl set-hostname $(hostname)")
 
+    # FIXME: work around for BZ 2084118 - Downgrading nmstate
```
michalskrivanek, line 122: as a workaround this would be likely in setup_host.sh

ost_utils/pytest/fixtures/deployment.py
```diff
@@ -118,6 +119,14 @@ def deploy(
...
+    if ansible_host0_facts.get("ansible_distribution_major_version") == "9":
+        ansible_hosts.shell("rpm -e --nodeps nmstate")
```
michalskrivanek, line 124: doing it in setup_host would also make it easier to handle node properly. we want to do this even in mixed mode of el9 host0 and el8 node host-1

ost_utils/pytest/fixtures/deployment.py
```diff
@@ -118,6 +119,14 @@ def deploy(
...
+        ansible_hosts.shell(
+            "dnf install -y \
```
michalskrivanek, line 126: and then you need to add it to excludes everywhere since we do dnf update here or there...

common/deploy-scripts/setup_host.sh
```diff
@@ -22,6 +22,16 @@ if [[ $(which nodectl) ]]; then
     echo 3 > /proc/sys/vm/drop_caches
 fi
 
+# FIXME: work around for BZ 2084118 - Downgrading nmstate
+DISTRO=$( cat /etc/redhat-release )
+if [[ "$DISTRO" == *"9"* ]]; then
+    rpm -e --nodeps nmstate
+    dnf config-manager --set-enabled appstream
+    dnf install -y http://mirror.stream.centos.org/9-stream/AppStream/x86_64/os/Packages/nmstate-2.0.0-0.4.alpha3.el9.noarch.rpm
+    dnf update --exclude="nmstate nmstate-plugin-ovsdb python3-libnmstate"
```
tinez, line 31: What do we need this update for?

hbraha: you're right not really needed, was for testing purposes

tinez: Now you've replaced it with `dnf --exclude="nmstate nmstate-plugin-ovsdb python3-libnmstate"` - this will fail - you didn't even pass any verb to dnf.

hbraha: you're right, so i didn't find any preferable verb to pass to dnf, and if we want to avoid `dnf update` so the other way is to exclude it directly at the dnf.conf

michalskrivanek:
but it's not really specific to OST - the right place to exclude it also for potential el9 users is ovirt-release rpm

tinez:
> but it's not really specific to OST - the right place to exclude it also for potential el9 users is ovirt-release rpm

Do we ship nmstate from oVirt's repos?

tinez:
> as far as i can see it ships from appstream:

Then we would either need to modify the appstream repo file or global dnf conf like we do here. Unless there are other ways to do it globally that I'm not aware of. Is it that important now? Don't we expect a fix for this soon?

michalskrivanek:
eh, ok, let's leave it in setup then. but why enable/disable appstream repo, @hbraha ?

hbraha:
> eh, ok, let's leave it in setup then. but why enable/disable appstream repo, @hbraha ?

since disable the repos in the ost runs, we need the appsream repo enabled in order to install the correct dependencies for the downgraded nmstate version.

hbraha:
> > > eh, ok, let's leave it in setup then. but why enable/disable appstream repo, @hbraha ?
> > 
> > 
> > since disable the repos in the ost runs, we need the appsream repo enabled in order to install the correct dependencies for the downgraded nmstate version.
> 
> but then does it not work to install the concrete version without using full URL(i.e. doing localinstall), if that version is available in appstream?

ack, changed.

PR #154, opened 2022-05-16
Try to workaround specific WebDriverException 

ljelinkova:
If the WebDriverException is caused by "_Expected to read a START_MAP but instead have: END. Last 0 characters read_" lets ignore it and try again.

If the TimeoutException is caused by "_Unable to execute request for an existing session: java.util.concurrent.TimeoutException_" lets ignore it and try again.

PR #163, opened 2022-05-24
Add ignoring of TimeoutException also to _wait_until

ljelinkova:
This is a follow-up patch to https://github.com/oVirt/ovirt-system-tests/pull/154. I am adding the comments from that patch as a separate comments to this patch.

### DIFF
--- a/ost_utils/selenium/navigation/driver.py
+++ b/ost_utils/selenium/navigation/driver.py
@@ -116,7 +116,9 @@ class Driver:
         self._wait_until(message, assert_utils.LONG_TIMEOUT, condition_method, *args)
 
     def _wait_until(self, message, timeout, condition_method, *args):
-        WebDriverWait(self.driver, timeout).until(ConditionClass(condition_method, *args), message)
+        WebDriverWait(self.driver, timeout, ignored_exceptions=[TimeoutException]).until(
+            ConditionClass(condition_method, *args), message
+        )
 
     def wait_while(self, message, condition_method, *args):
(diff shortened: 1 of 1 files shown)


ljelinkova:
_Don't we want the WebDriverException on that ignored_exceptions too?_ (https://github.com/oVirt/ovirt-system-tests/pull/154#discussion_r877032210)

ljelinkova:
I'd rather use f-strings throughout the code - they're much more elegant (https://github.com/oVirt/ovirt-system-tests/pull/154#discussion_r877034298)

ljelinkova:
Let's use PEP8 conventions, so should_run_again (https://github.com/oVirt/ovirt-system-tests/pull/154#discussion_r877034958)

ljelinkova:
Can other situations than java.util.concurrent.TimeoutException indeed cause a python TimeoutException to be raised? (https://github.com/oVirt/ovirt-system-tests/pull/154#discussion_r877039046)

ost_utils/selenium/navigation/driver.py
```diff
@@ -155,7 +157,7 @@ def __call__(self, driver):
             raise e
         except Exception as e:
             LOGGER.exception(
-                '!!!ConditionClass failed with ' + e.__class__.__name__ + ' at retry number ' + str(self.retry)
+                f'!!! ConditionClass failed with {e.__class__.__name__} at retry number {str(self.retry)}'
```
tinez, line 160: We don't need the `str()` conversions around `self.retry`

PR #166, opened 2022-05-26
el9stream: Reboot all hosts

tinez:
On el9stream hosts we configure libvirt to use the monolithic daemon
during host deployment [1]. Until now we used to reboot only one of the
hosts in a basic suite run to make it faster, but that libvirt change
requires all hosts to be rebooted.

[1] https://github.com/oVirt/ovirt-engine/pull/80

### DIFF
--- a/ost_utils/pytest/fixtures/defaults.py
+++ b/ost_utils/pytest/fixtures/defaults.py
@@ -28,8 +28,11 @@ def hostnames_to_add(hosts_hostnames):
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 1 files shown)


michalskrivanek (approved):
eh, ok. thanks!

tinez:
Ran OST 6 times with that - none of the runs failed.

michalskrivanek:
we need to install the images too...and then it should work!:)

tinez:
They are installed on some executors, but not all of them.

PR #170, opened 2022-05-30
[UI] add a test for db connection in grafana

avlitman:
Jira-ticket:https://issues.redhat.com/browse/RHV-42170


avlitman:
Verified on both el8stream and rhel8, the new test pass as expected and fail as expected.

basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
```diff
@@ -588,6 +590,19 @@ def test_grafana(
     grafana.wait_for_displayed()
     save_screenshot('grafana')
 
+    # navigate directly to Grafana Configuration/Data Sources page
+    ovirt_driver.driver.get("https://"+engine_fqdn+"/ovirt-engine-grafana/datasources")
+    ovirt_driver.xpath_wait_and_click('Open oVirt DWH Datasource', '//*[text()="oVirt DWH"]')
+    ovirt_driver.xpath_wait_and_click('Save & Test button', '//*[text()="Save & Test"]')
+    # make sure that after clicking Save & Test button, "Database Connection OK" is popping up
+    time.sleep(5)
```
avlitman, line 598: time.sleep(1) is also enough.

avlitman: > ok, so check it immediately 
I checked without waiting at all, in firefox it works but in chrome it failed.

>and loop 10x with a 1s delay and only then fail. we do not want "mandatory" sleeps...
I have tried other methods like WebDriverWait but I could not make it work (yet).
Not sure what you mean by 'loop 10x with a 1s delay', maybe there is a similar example that already exists in the code?

basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
```diff
@@ -588,6 +590,19 @@ def test_grafana(
     grafana.wait_for_displayed()
     save_screenshot('grafana')
 
+    # navigate directly to Grafana Configuration/Data Sources page
+    ovirt_driver.driver.get("https://"+engine_fqdn+"/ovirt-engine-grafana/datasources")
```
tinez, line 594: Please use f-strings instead of concatenating with `+`

basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
```diff
@@ -21,6 +21,7 @@
 
 from selenium import webdriver
 from selenium.webdriver.common.desired_capabilities import DesiredCapabilities
+from selenium.webdriver.common.by import By
```
tinez, line 24: This import should go before `desired_capabilities` one

basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
```diff
@@ -588,6 +590,19 @@ def test_grafana(
...
 
+    # navigate directly to Grafana Configuration/Data Sources page
+    ovirt_driver.driver.get("https://"+engine_fqdn+"/ovirt-engine-grafana/datasources")
+    ovirt_driver.xpath_wait_and_click('Open oVirt DWH Datasource', '//*[text()="oVirt DWH"]')
+    ovirt_driver.xpath_wait_and_click('Save & Test button', '//*[text()="Save & Test"]')
+    # make sure that after clicking Save & Test button, "Database Connection OK" is popping up
+    time.sleep(5)
+    ovirt_driver.driver.find_element(
+        By.XPATH,
+        '//grafana-app/div/div/react-container/div/div/div[1]/div/div[2]/div/form/div[3]/div/div[2]/div[text()="Database Connection OK"]',
```
tinez, line 601: This looks way too complicated, isn't there an element on the path we can grab with an id?

basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
tinez: I'd rather not reimplement retry logic here - please consult with @ljelinkova and let's get that WebDriver methods working.

basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
avlitman: Grafana do not use ids nor meaningful classes:
Screenshot from 2022-06-15 11-02-47
 
I will try to make it more simple with @ljelinkova

tinez: Hmm ok. Can we use selectors to pick those aria-labels?

ljelinkova (changes requested):

basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
```diff
@@ -588,6 +592,21 @@ def test_grafana(
     grafana.wait_for_displayed()
     save_screenshot('grafana')
 
+    # navigate directly to Grafana Configuration/Data Sources page
+    ovirt_driver.driver.get(f"https://{engine_fqdn}/ovirt-engine-grafana/datasources")
+    ovirt_driver.xpath_wait_and_click('Open oVirt DWH Datasource', '//*[text()="oVirt DWH"]')
+    ovirt_driver.xpath_wait_and_click('Save & Test button', '//*[text()="Save & Test"]')
+    # make sure that after clicking Save & Test button, "Database Connection OK" is popping up
+    WebDriverWait(ovirt_driver.driver, 5).until(
```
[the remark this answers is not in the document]
avlitman: I see we also use WebDriverWait in the code, I will need to investigate more about wait_until since I was not able to make it work yet.

basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
```diff
@@ -588,6 +592,21 @@ def test_grafana(
...
+        EC.presence_of_element_located(
+            (
+                By.XPATH,
+                '//*[contains(@aria-label, "Data source settings page Alert")]//*[text()="Database Connection OK"]',
+            )
+        )
+    )
```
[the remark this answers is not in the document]
avlitman: The issue is that the "Database Connection OK" is visible only after this steps:
1.    ovirt_driver.xpath_wait_and_click('Open oVirt DWH Datasource', '//*[text()="oVirt DWH"]')
2.    ovirt_driver.xpath_wait_and_click('Save & Test button', '//*[text()="Save & Test"]')
and we agreed that step 1 should be in the test file so that it is more visible that we are navigating directly and not using the menu.

basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
```diff
@@ -588,6 +592,21 @@ def test_grafana(
     grafana.wait_for_displayed()
     save_screenshot('grafana')
 
+    # navigate directly to Grafana Configuration/Data Sources page
+    ovirt_driver.driver.get(f"https://{engine_fqdn}/ovirt-engine-grafana/datasources")
```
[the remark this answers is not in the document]
avlitman: > Please use single quotes - just for the consistency with other strings.

We use double quotes in most places, e.g. https://github.com/oVirt/ovirt-system-tests/blob/b6b822d40f814e132cb562be3abbd210e5fff227/ost_utils/pytest/fixtures/keycloak.py#L12

basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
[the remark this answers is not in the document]
ljelinkova: it is used only by `wait_until` and `wait_while` in the `driver.py`. The signature of those method is:
- text description of the wait that will be part of the exception if the wait fails
- method that returns true/false (usually some is_present method)
- arguments of that method

Something like this should work:
```
ovirt_driver.wait_until('"Database Connection OK" string is present', ovirt_driver.is_xpath_present, '//*[contains(@aria-label, "Data source settings page Alert")]//*[text()="Database Connection OK"]')
```

basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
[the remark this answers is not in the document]
ljelinkova: We agreed that the loading of the page would stay in the test file. I would keep the following in the test:
```
    # navigate directly to Grafana Configuration/Data Sources page
    ovirt_driver.driver.get(f"https://{engine_fqdn}/ovirt-engine-grafana/datasources")
```
And moved the rest to the Grafana.py
```
    ovirt_driver.xpath_wait_and_click('Open oVirt DWH Datasource', '//*[text()="oVirt DWH"]')
    ovirt_driver.xpath_wait_and_click('Save & Test button', '//*[text()="Save & Test"]')
    # make sure that after clicking Save & Test button, "Database Connection OK" is popping up
    WebDriverWait(ovirt_driver.driver, 5).until(
        EC.presence_of_element_located(
            (
                By.XPATH,
                '//*[contains(@aria-label, "Data source settings page Alert")]//*[text()="Database Connection OK"]',
            )
        )
    )
```

ost_utils/selenium/page_objects/Grafana.py
```diff
@@ -23,6 +23,20 @@ def is_displayed(self):
     def get_displayable_name(self):
         return 'Grafana'
 
+    def testConnection(self):
+        self.ovirt_driver.xpath_wait_and_click('Open oVirt DWH Datasource', '//*[text()="oVirt DWH"]')
+        self.ovirt_driver.xpath_wait_and_click('Save & Test button', '//*[text()="Save & Test"]')
+        # make sure that after clicking Save & Test button, "Database Connection OK" is popping up
+        try:
+            self.ovirt_driver.wait_until(
```
avlitman, line 31: @ljelinkova 
So it's working, but takes much longer to fail (~205s) than WebDriverWait (29s).

ljelinkova: That is expected - it uses `assert_utils.SHORT_TIMEOUT` from `ost_utils/assert_utils.py` which is set to 180s, whereas your previous code used only 5s timeout.

ljelinkova (approved):

ost_utils/selenium/page_objects/Grafana.py
avlitman: > That is expected - it uses `assert_utils.SHORT_TIMEOUT` from `ost_utils/assert_utils.py` which is set to 180s, whereas your previous code used only 5s timeout.

+1 Thanks

ost_utils/selenium/page_objects/Grafana.py
```diff
@@ -23,6 +23,24 @@ def is_displayed(self):
     def get_displayable_name(self):
         return 'Grafana'
 
+    def testConnection(self):
```
tinez, line 26: Please use `this_convention` for function names. `test_connection` is a bad idea as pytest might pick it up as a test :)

ost_utils/selenium/page_objects/Grafana.py
```diff
@@ -23,6 +23,24 @@ def is_displayed(self):
...
+        self.ovirt_driver.xpath_wait_and_click('Open oVirt DWH Datasource', '//*[text()="oVirt DWH"]')
+        self.ovirt_driver.xpath_wait_and_click('Save & Test button', '//*[text()="Save & Test"]')
+        # make sure that after clicking Save & Test button, "Database Connection OK" is popping up
+        try:
```
tinez, line 30: So I'm actually a bit confused by the logic below now. What happens when that `self.ovirt_driver.wait_until` call times out? Do we get an exception? What kind of exception?

Catching `except Exception` is not that good here - for any problem that might occur during this call we will say "it's because that string is not present". That's not necessarily true. Can we limit the exception type here?

tinez (changes requested):

ost_utils/selenium/page_objects/Grafana.py
avlitman: > So I'm actually a bit confused by the logic below now. What happens when that `self.ovirt_driver.wait_until` call times out? Do we get an exception? What kind of exception?

timeout exception with the note:
"Database Connection OK" string is not present, This may mean that the Grafana UI has changed and we will no longer be able to detect the string be located"

> Catching `except Exception` is not that good here - for any problem that might occur during this call we will say "it's because that string is not present". That's not necessarily true. Can we limit the exception type here?

I did limit to 5 secondes, but Lucie asked me to use this method instead:
"Please use ovirt_driver.wait_until as it is used in other parts of the code. That method sets default timeouts and has additional fix for TimeoutExceptions."
see resolved conversations.

avlitman: In addition I verified that the new test pass as expected and fail as expected.
without the timeout limit it will fail in ~200sec and with ~30sec.

avlitman: > So I'm actually a bit confused by the logic below now. What happens when that `self.ovirt_driver.wait_until` call times out? Do we get an exception? What kind of exception?

We get a timeout exception, it look like that:
Screenshot from 2022-06-21 11-27-11

> Catching `except Exception` is not that good here - for any problem that might occur during this call we will say "it's because that string is not present". That's not necessarily true. Can we limit the exception type here?

I used 5 sec timmout at first, but Lucia asked me to use this specific method:
"Please use ovirt_driver.wait_until as it is used in other parts of the code. That method sets default timeouts and has additional fix for TimeoutExceptions." (see resolved conversations).

tinez (approved):

PR #177, opened 2022-06-03
el9stream: Use el9-based node

tinez:


michalskrivanek:
vdsm numbering is broken, so it won't work...but it should after nightly node rebuild....

PR #178, opened 2022-06-06
Disable keycloak if it's not enabled

didib:
By passing 'None', we let stuff rely on defaults. Better be explicit.

Change-Id: I03fc6e39e094dc73a348e581c59d87e18bf18ee5


arso (approved):

michalskrivanek:
not ideal as we said that the default on rhel would be "without keycloak", we broke it, apparently.
but alright, let's reenable HE suite

michalskrivanek:
oh ok, we need ovirt-hosted-engine-setup-2.6.4 too

didib:
Fails on rhel8 because we are missing there a build of ovirt-hosted-engine-setup 2.6.4.

michalskrivanek:
failed, but just on a flaky test towards the end. keycloak stuff worked

PR #180, opened 2022-06-07
README: Note about permissions

didib:
Without group 'libvirt', virsh says:

    system policy prevents management of local virtualized systems

Change-Id: I66727d641244698ed1b702f5d2fd2415f890d6be


tinez:
`./setup_for_ost.sh` already sets up the appropriate permissions both in single- and multi-user configs. We already mention it in the `README.md`, so I wouldn't introduce this change not to cause confusion.

tinez:
`./setup_for_ost.sh` already sets up the appropriate permissions both in single- and multi-user configs. We already mention it in the `README.md`, so I wouldn't introduce this change not to cause confusion.

PR #185, opened 2022-06-08
Align SNMP config and testing with V3

emesika:
This patch tests ovirt-notification-service with SNMP V3 security level 3.
All configuration files were changed to work with that.
The test is looking for VDC_USER_LOGIN trap that must be in /var/log/snmptrapd.log
since the ovirt-engine-notifier service is started before ovirt-engine.

Signed-off-by: Eli Mesika <emesika@redhat.com>


mwperina (approved):
+1

common/deploy-scripts/setup_engine.sh
```diff
@@ -19,19 +19,61 @@ cat > /etc/cron.d/ovirt-engine << EOF
...
 
 cp /usr/share/doc/ovirt-engine/mibs/* /usr/share/snmp/mibs
 
+systemctl start snmpd
 systemctl start snmptrapd
+# new user created in net-snmp/snmpd*.conf are available only after services are restarted
+# TODO , check if this is a bug
+systemctl restart snmpd
+systemctl restart snmptrapd
+systemctl enable snmpd
```
tinez, line 76: I'm not sure whether we need the `enable` verbs here - we never really restart the engine VM. Maybe we could remove these and call `systemctl enable --now snmpd` etc. instead of doing `systemctl start`?

common/deploy-scripts/setup_engine.sh
```diff
@@ -19,19 +19,61 @@ cat > /etc/cron.d/ovirt-engine << EOF
...
+systemctl stop snmpd
+systemctl stop snmptrapd
+cat > /etc/snmp/snmptrapd.conf << EOF
+disableAuthorization yes
+# version 3 traps: allow user ovirtengine to log,execute,net
+authUser log,execute,net ovirtengine
+# version 3 add a user NoAuthnoPriv who can send noAuthNoPriv
+authUser log,execute,net NoAuthNoPriv noauth
+# Log incoming traps to /var/log/snmptrapd.log
+[snmp] logOption f /var/log/snmptrapd.log
```
tinez, line 50: This is a really weird syntax for a conf file... that `[snmp]` part is a beginning of a new section? If so, maybe we should put the following text after a newline?

common/deploy-scripts/setup_engine.sh
```diff
@@ -19,19 +19,61 @@ cat > /etc/cron.d/ovirt-engine << EOF
...
+SNMP_PRIVACY_PASSPHRASE=privpass
+SNMP_SECURITY_LEVEL=3
 EOF
 chown ovirt:ovirt /etc/ovirt-engine/notifier/notifier.conf.d/20-snmp.conf
 
-echo "[snmp] logOption f /var/log/snmptrapd.log" >> /etc/snmp/snmptrapd.conf
-echo "disableAuthorization yes" >> /etc/snmp/snmptrapd.conf
+# Create /var/log/snmptrapd.log with correct SELinux file type
+touch /var/log/snmptrapd.log
+chcon -t snmpd_log_t /var/log/snmptrapd.log
```
tinez, line 38: Does `snmptrapd` create that log file with wrong context? Wouldn't that be a bug?

common/deploy-scripts/setup_engine.sh
emesika: well , start is not working and does not really create the user , the solution is currently to restart after start , the 'enable' was there before my patch , I can remove it , but since this is only a test , I don't know if it worth

common/deploy-scripts/setup_engine.sh
emesika: This is the syntax needed for this entry to work , newline will not work here

common/deploy-scripts/setup_engine.sh
emesika: I will try to reproduce and open a BZ if needed , anyway ' this will stay as a workaround

PR #186, opened 2022-06-13
Check VM Portal status based on presence of Refresh button

rszwajko:
After migrating VM Portal to PatternFly4 the CSS class used by the
spinner component has changed to "pf-c-spinner".

The new class should be checked together with the old class to ensure
smooth transition from PF3 to PF4.


rszwajko:
Converted to draft - let's refactor `is_displayed` to use  `appConfigured` flag.

michalskrivanek:
still relevant with #224 in?

rszwajko:
> still relevant with #224 in?

Yes. #224 fixes problems with ui-extensions.

PR #187, opened 2022-06-14
WIP: he: Log HostedEngine VM console

didib:
Requires https://github.com/oVirt/vdsm/pull/249 .

TODO: If we want this, once above is merged, we should probably do this
using python code, and on both hosts. If HE VM fails to start on host-0,
HA on host-1 might try to start it, and we want logging there as well.

Change-Id: If3b176d151495debaa86de0be6fdb95e016bd30c


michalskrivanek:
we should install it in ost-images, but then...you can as well just use setup_host.sh to add it to both hosts.
I wouldn't add anything more to the the initialization fixtures...that ansible there is so horribly slow and awful to debug....IMHO "static" initialization/customization is best in those setup_ *scrips.

common/deploy-scripts/setup_first_he_host.sh
```diff
@@ -154,6 +154,12 @@ ip -6 -o addr show dev eth0 scope global | grep -q eth0 && fix_ipv6
 # and if the package is excluded it fails, we need to convince DNF that current version is good enough.
 dnf config-manager --setopt=best=False --save
 
+# Requires https://github.com/oVirt/vdsm/pull/249
+echo 'log_console_vm_regexp=HostedEngine' > /etc/sysconfig/vdsm
+# TODO: I am using this right now with a custom repo, so already enabled. Fix if we eventually want to merge.
```
michalskrivanek, line 159: I think you can drop teh comments, it's in copr already, but it unfortunately missed d/s vdsm this week...either skip on rhel and let's merge this now or wait with this pr for few more days

didib: IMO we want logging also on host-1, so above won't be enough

michalskrivanek: setup_host.sh would be best then

common/deploy-scripts/setup_first_he_host.sh
```diff
@@ -154,6 +154,12 @@ ip -6 -o addr show dev eth0 scope global | grep -q eth0 && fix_ipv6
...
+# Probably by adding to ost-images.
+dnf install -y vdsm-hook-log-console
```
michalskrivanek, line 161: https://github.com/oVirt/ost-images/pull/66 merged

didib:
https://redir.apps.ovirt.org/dj/job/ds-ost-baremetal_manual/41102/

michalskrivanek (approved):

PR #188, opened 2022-06-14
Workaround for metrics configuration ansible playbook hanging forever

tinez:
We're experiencing situations where the ansible playbook for configuring
metrics is hanging forever. This problem has been analyzed and it's most
probably an internal issue of either ansible-core or python interpreter. Since
it only seems to be an OST problem let's add a timeout and retries for this.

### DIFF
--- a/ost_utils/utils.py
+++ b/ost_utils/utils.py
@@ -11,8 +11,24 @@ import os
 import queue
 import sys
 import threading
+import time
 
 LOGGER = logging.getLogger(__name__)
+
+
+class EggTimer:
+    def __init__(self, timeout):
+        self.timeout = timeout
+
+    def __enter__(self):
+        self.start_time = time.time()
+        return self
+
+    def __exit__(self, *_):
+        pass
+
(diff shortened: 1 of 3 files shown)


ost_utils/utils.py
```diff
@@ -31,33 +58,69 @@ def func_vector(target, args_sequence):
...
+
+    def _gather_results(self):
+        for q in self.queues:
+            self.results.append(q.get(block=False))
+
+    def _handle_exceptions(self, raise_exceptions):
+        exceptions = [result['exception'] for result in self.results if 'exception' in result]
+
+        if exceptions:
+            LOGGER.debug(f"{len(exceptions)} out of {len(self.targets)} threads raised exceptions")
```
acernek, line 120: Maybe we could log the exceptions? Raising just the first one might prolong debugging etc.

tinez: Sure, good idea

acernek (approved):

PR #191, opened 2022-06-15
basic: enable graphics test on el9

michalskrivanek:
With https://github.com/oVirt/ovirt-engine/pull/75 we no longer need to
skip this test on el9, VNC graphics is selected by default and allows
vm0 to run later on


michalskrivanek:
el9 testes manually

PR #194, opened 2022-06-20
el9: Skip test broken by libvirt regression

tinez:


michalskrivanek (approved):
yeah, I think test_next_run_unplug_cpu takes care of the actually-unplugged nic by restarting the vm so this indeed should work

tinez:
The other test uses that NIC most probably, let's try with 2 of them disabled

tinez:
Works. Unfortunately we cannot use `strict` that would fail the test once the issue is fixed. Let's post a revert PR and check on this in a week.

PR #196, opened 2022-06-21
ost.sh: Add fetch-artifacts command

acernek:
While fixtures for collecting artifacts are very handy and work
really well, it's useful to have a tool for adhoc artifact
downloading to cover cases of OST runs that are hanging forever
in some place.

### DIFF
--- a/common/scripts/fetch_artifacts.yml
+++ b/common/scripts/fetch_artifacts.yml
@@ -1,0 +1,58 @@
+#!/usr/bin/ansible-playbook
+
+- hosts: all
+
+  vars:
+    archive_name: artifacts.tar.gz
+    remote_archive_path: "/var/tmp/{{ archive_name }}"
+    artifact_list:
+      - /etc/dnf
+      - /etc/firewalld
+      - /etc/grafana
+      - /etc/httpd/conf
+      - /etc/httpd/conf.d
+      - /etc/httpd/conf.modules.d
+      - /etc/ovirt-engine
+      - /etc/ovirt-engine-dwh
+      - /etc/ovirt-engine-metrics
+      - /etc/ovirt-engine-setup.conf.d
+      - /etc/ovirt-engine-setup.env.d
+      - /etc/ovirt-host-deploy.conf.d
+      - /etc/ovirt-imageio-proxy
+      - /etc/ovirt-provider-ovn
+      - /etc/ovirt-vmconsole
+      - /etc/ovirt-web-ui
+      - /etc/resolv.conf
+      - /etc/sysconfig
+      - /etc/yum
+      - /etc/yum.repos.d
+      - /root
+      - /tmp/dnf_yum.conf
+      - /var/cache/ovirt-engine
+      - /var/lib/ovirt-engine/setup/answers
+      - /var/lib/ovirt-engine/ansible-runner
+      - /var/lib/pgsql/initdb_postgresql.log
+      - /var/lib/pgsql/data/log
+      - /var/log
+    ost_repo_root: "{{ lookup('env', 'OST_REPO_ROOT') }}"
+
+  tasks:
+    - name: Get the journal right before collecting
+      ansible.builtin.shell: journalctl -a --no-pager -o short-iso-precise > /var/log/journalctl.log
+
+    - name: Archive artifacts
+      community.general.archive:
+        path: "{{ artifact_list }}"
+        dest: "{{ remote_archive_path }}"
+
+    - name: Fetch archived artifacts
+      ansible.builtin.fetch:
+        src: "{{ remote_archive_path }}"
+        dest: "{{ ost_repo_root }}/exported-artifacts/test_logs/{{ inventory_hostname }}/{{ archive_name }}"
+        flat: yes
+
+    - name: Extract the artifacts
+      local_action: ansible.builtin.command tar -xf {{ ost_repo_root }}/exported-artifacts/test_logs/{{ inventory_hostname }}/{{ archive_name }} -C {{ ost_repo_root }}/exported-artifacts/test_logs/{{ inventory_hostname }}
+
+    - name: Clean up the archives
+      local_action: ansible.builtin.command rm {{ ost_repo_root }}/exported-artifacts/test_logs/{{ inventory_hostname }}/{{ archive_name }}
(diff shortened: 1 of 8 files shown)


lagofy.sh
```diff
@@ -312,6 +312,20 @@ ost_console() {
...
 }
 
+ost_fetch_artifacts() {
+    _deployment_exists || return 1
+    mkdir -p "${OST_REPO_ROOT}/exported-artifacts"
+    ansible-playbook -i deployment/ansible_inventory/backend.yml -u root --ssh-common-args '-o StrictHostKeyChecking=no' common/scripts/fetch_artifacts.yml >/dev/null || {
+        echo "ansible playbook failed"
+        return 9
+    };
+    sed -n "/^ost/ s/ansible[a-z_]*=//g p" $OST_DEPLOYMENT/hosts | while IFS=\  read -r host ip key; do
```
michalskrivanek, line 322: is it then worth using ansible for tarring the files when you anyway use the hosts file directly here?

michalskrivanek: we can as well just scp whatever we need directly without the horrible ansible overhead

michalskrivanek (commented):
I would expect the fixture to be modified to use this?

lagofy.sh
tinez: scp doesn't support copying directories. We'd need to enumerate all the files by ourselves.

lagofy.sh
```diff
@@ -312,6 +312,20 @@ ost_console() {
     fi
 }
 
+ost_fetch_artifacts() {
+    _deployment_exists || return 1
+    mkdir -p "${OST_REPO_ROOT}/exported-artifacts"
+    ansible-playbook -i deployment/ansible_inventory/backend.yml -u root --ssh-common-args '-o StrictHostKeyChecking=no' common/scripts/fetch_artifacts.yml >/dev/null || {
```
tinez, line 318: Please use `deployment/ansible_inventory` directory as the inventory - `backend.yml` doesn't have information on HE.

common/scripts/fetch_artifacts.yml
```diff
@@ -0,0 +1,50 @@
...
+    remote_archive_path: "/var/tmp/{{ archive_name }}"
+
+  tasks:
+    - name: Get the journal right before collecting
+      ansible.builtin.shell: journalctl -a --no-pager -o short-iso-precise > /var/log/journalctl.log
+
+    - name: Archive artifacts
+      community.general.archive:
+        path:
+          - /etc/dnf
```
tinez, line 16: If you could find a way of avoiding the duplication between this file list and the one in `artifacts.py` it would be great.

acernek: Done

lagofy.sh
acernek: Done

lagofy.sh
```diff
@@ -312,6 +312,20 @@ ost_console() {
     fi
 }
 
+ost_fetch_artifacts() {
+    _deployment_exists || return 1
+    mkdir -p "${OST_REPO_ROOT}/exported-artifacts"
```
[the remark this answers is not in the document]
acernek: 'Should' is not enough IMO, better safe than sorry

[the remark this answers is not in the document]
michalskrivanek: oh yeah, let's check for NULL on every access of every object, welcome to ovirt-engine!
no
there's going to be only one place where we create each resource.

lagofy.sh
michalskrivanek: -r      Recursively copy entire directories.  Note that scp follows symbolic links encountered in the tree traversal.

?

ost_utils/pytest/fixtures/artifacts.py
```diff
@@ -25,36 +26,9 @@ def artifacts_dir():
...
-        '/root',
-        '/tmp/dnf_yum.conf',
-        '/var/cache/ovirt-engine',
-        '/var/lib/ovirt-engine/setup/answers',
-        '/var/lib/ovirt-engine/ansible-runner',
-        '/var/lib/pgsql/initdb_postgresql.log',
-        '/var/lib/pgsql/data/log',
-        '/var/log',
-    ]
+    with open(os.path.join(os.environ["OST_REPO_ROOT"], "common", "scripts", "fetch_artifacts.yml"), "r") as f:
```
michalskrivanek, line 29: the python side of collection should be somplified. you do not need the list when you have a playbook that processes everything. just call it and be done. Best to just call ost.sh fetch-artifacts.

lagofy.sh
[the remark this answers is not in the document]
acernek: Ansible handles it anyway, so not needed at all

ost_utils/pytest/fixtures/artifacts.py
acernek: Done

tinez (commented):
Found 2 nits, but overall I really like the result.

lagofy.sh
```diff
@@ -312,6 +312,14 @@ ost_console() {
     fi
 }
 
+ost_fetch_artifacts() {
+    _deployment_exists
+    ansible-playbook -i deployment/ansible_inventory -u root --ssh-common-args '-o StrictHostKeyChecking=no' common/scripts/fetch_artifacts.yml >/dev/null || {
```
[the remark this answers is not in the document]
michalskrivanek: yes please

ost_utils/pytest/fixtures/artifacts.py
```diff
@@ -23,71 +23,14 @@ def artifacts_dir():
...
-    calls = [
-        functools.partial(collect, hostname, artifact_list, artifacts_dir)
-        for hostname, artifact_list in artifacts.items()
-    ]
-    utils.invoke_different_funcs_in_parallel(*calls)
+    playbook = os.path.join(root_dir, "common", "scripts", "fetch_artifacts.yml")
+    shell.shell(
+        f"ansible-playbook -i {ansible_inventory.dir} -u root --ssh-common-args '-o StrictHostKeyChecking=no' {playbook}",
+        shell=True,
+    )
```
michalskrivanek, line 33: when you're calling out to shell...it makes sense to keep only one implementation, just call our to ost.sh fetch-artifacts instead.

lagofy.sh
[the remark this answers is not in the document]
acernek: Done (I thought the userknownhostsfile is already set in the inventory, but no hurt in doing it here as well)

tinez (approved):

PR #198, opened 2022-06-21
Test that notifier works after restart

emesika:
Verify that snmp traps works after restart

This patch verifies that SNMP traps are still working after
ovirt-engine-notifier and snmptrapd services are restarted.

First the services are restarted, then a new tag "notifiertag" is added to
the system, finally we check that a trap generated on the new tag created is
logged to snmptrapd.log

We have to set the configuration value DAYS_TO_SEND_ON_STARTUP to 1 in engine setup
because its default value is 0, it means that it will not send any old events,
a race between the service start and the new "notifiertag" tag will cause that this
event will not be processed.


mwperina (approved):
+1

michalskrivanek (approved):

PR #199, opened 2022-06-21
he: increase host-0 memory 1GB

michalskrivanek:
sometimes HE VM has troubles booting and it seeems to be related to
memory fragmentation or some other problem during initial allocation.
Let's try with 1GB more on host-0 only, as once it boots there is no
need for so much memory


michalskrivanek:
seems it's rather #204

PR #201, opened 2022-06-22
setup_for_ost: Move collection install to a playbook

acernek:


acernek:
Works on clean centos stream 8 VM

tinez (approved):

PR #207, opened 2022-06-29
Log firmware

didib:


didib:
Closing, for now. If we want to merge it to always log the firmware - might be useful for times where some infra - e.g. qemu/seabios/libvirt is unstable - we should merge https://github.com/oVirt/vdsm/pull/271 and then update current.

PR #208, opened 2022-06-30
Fix ovirt admin email to work with grafana

arso:

PR #210, opened 2022-07-05
Revert "Run DB connection test for upstream only"

avlitman:
Reverts oVirt/ovirt-system-tests#202 since we fixed the issue here: https://github.com/oVirt/ovirt-system-tests/pull/208


tinez:
We cannot revert that PR - the UI between the u/s and d/s Grafana versions will stay different. Sorry for confusion.

tinez (approved):

PR #212, opened 2022-07-07
UI: remove test_add_grafana_user method

avlitman:
We are using oVirt Engine Auth to connect to Grfana,
seems like there is no need in this method.


mwperina (approved):
Let's remove code which is not used. And great we have a confirmation that

1. SSO within Grafana really works (I had a suspicion that this internal Grafana with the same email as engine admin user could interfere with SSO)
2. https://github.com/oVirt/ovirt-system-tests/pull/208/files is not needed for AAA to work with Grafana

avlitman:
> 2. https://github.com/oVirt/ovirt-system-tests/pull/208/files is not needed for AAA to work with Grafana

so we can close #208 and merge #210 Right?

PR #213, opened 2022-07-07
ovirtlib: Use private_data_dir in Playbook run

acernek:
When you don't pass the 'private_data_dir' argument to
'ansible_runner.run', ansible-runner creates a dir by itself
and never cleans it up. The result is that CI agents have
'/tmp' dirs flooded with ansible logs.


tinez (approved):

network-suite-master/ovirtlib/ansiblelib.py
```diff
@@ -38,12 +37,13 @@ def _run_playbook_executor(self):
             playbook=self._playbook,
             extravars=self._extra_vars,
             inventory='localhost ansible_connection=local',
+            private_data_dir=pd.PrivateDir.get(),
```
erav, line 40: line 40 creates a static dependency of ovirtlib on external ost module. 
This inhibits testing the Playbook class easily or using ovirtlib in other projects.
It also conceals the true dependencies of Playbook, forcing the class user to read its internal code.
A better option would be to inject a private dir path into the Playbook constructor when the instance is created in client code. This should solve all above problems.
So please create a fixture in network-suite-master/fixtures/ansible that returns a fresh private dir on each invocation and inject it into the Playbook instances where they are created.
The reason for the fixture is to insulate network test code from dependencies as well.
Thanks

network-suite-master/ovirtlib/ansiblelib.py
```diff
@@ -13,13 +13,14 @@ class AnsibleExecutionFailure(Exception):
     pass
 
 
-class Playbook(object):
-    def __init__(self, playbook, extra_vars=None):
+class Playbook:
+    def __init__(self, playbook, extra_vars=None, private_dir=None):
```
tinez, line 17: Do we have any paths that do not provide `private_dir`? If not I'd remove the default value.

network-suite-master/ovirtlib/ansiblelib.py
tinez: > So please create a fixture in network-suite-master/fixtures/ansible that returns a fresh private dir on each invocation

With current code we get 1 private dir per whole pytest session (`scope="session"`). I'm not sure whether this is what Eitan wanted, but there's one quirk with this approach - you cannot spawn multiple ansible-runner instances with the same private dir at the same time, so it won't work with i.e. multiple threads that do ansible stuff (we do have this scenarios in basic suite).

acernek: I'm aware, but was also not sure what Eitan meant - in your case it's not enough to change the scope, but also would need to return a factory instead if I understand it properly...

acernek: Eitan says he is fine with once per session

network-suite-master/ovirtlib/ansiblelib.py
acernek: Done

PR #214, opened 2022-07-07
check engine health after backup

emesika:
Add a health check after backup to delay tests until engine is healthy

### DIFF
--- a/basic-suite-master/test-scenarios/test_002_bootstrap.py
+++ b/basic-suite-master/test-scenarios/test_002_bootstrap.py
@@ -1113,7 +1113,7 @@ def test_verify_notifier(ansible_engine, ost_dc_name):
 
 
 @order_by(_TEST_LIST)
-def test_verify_engine_backup(ansible_engine, engine_api, ost_dc_name, is_node_suite):
+def test_verify_engine_backup(ansible_engine, engine_api, engine_fqdn, engine_download, ost_dc_name, is_node_suite):
     ansible_engine.file(path='/var/log/ost-engine-backup', state='directory', mode='0755')
 
     engine = engine_api.system_service()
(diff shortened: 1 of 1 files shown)


basic-suite-master/test-scenarios/test_002_bootstrap.py
```diff
@@ -1160,6 +1160,12 @@ def test_verify_engine_backup(ansible_engine, engine_api, ost_dc_name, is_node_s
         '--otopi-environment=OVESETUP_SYSTEM/memCheckEnabled=bool:False'
     )
 
+    # verify engine health after backup
+    url = 'https://{}/ovirt-engine/services/health'.format(engine_fqdn)
```
tinez, line 1164: We can use an f-string here.

basic-suite-master/test-scenarios/test_002_bootstrap.py
```diff
@@ -1160,6 +1160,12 @@ def test_verify_engine_backup(ansible_engine, engine_api, ost_dc_name, is_node_s
...
+    assert assert_utils.equals_within_short(
+        lambda: engine_download(url), b"DB Up!Welcome to Health Status!", [(shell.ShellError)]
```
tinez, line 1166: I'd add `allowed_exceptions=` before that shell error exception - it's better than trying to figure out what this parameter does. Also, we don't need `()` surrounding `shell.ShellError`.

tinez (commented):
Apart from minor nits, looks good. I'm tempted to remove that 10s sleep in `test_upload_cirros_image` that comes straight after engine backup... WDYT @bennyz ?

tinez:
@emesika could you please remove that sleep in this PR too?

basic-suite-master/test-scenarios/test_002_bootstrap.py
emesika: will do

PR #217, opened 2022-07-19
lagofy: Export ost-images package lists

acernek:
When installing an ost-image we also get .txt file with all the
packages installed inside. We want those lists to be available
in OST runs' `exported-artifacts` to make debugging issues
caused by a package update easier.


common/helpers/ost-images.sh
```diff
@@ -28,9 +28,11 @@ OST_IMAGES_SSH_KEY=${OST_IMAGES_SSH_KEY:-${OST_IMAGES_DIR}/${OST_IMAGES_DISTRO}_
 # FIXME: until we will build el9stream enigne image, we can use el8 engine with el9 hosts
 [[ "$OST_IMAGES_DISTRO" == "el9stream" ]] && OST_IMAGES_ENGINE_INSTALLED=${OST_IMAGES_ENGINE_INSTALLED/el9stream/el8stream}
 
-for i in OST_IMAGES_BASE OST_IMAGES_NODE OST_IMAGES_ENGINE_INSTALLED OST_IMAGES_HOST_INSTALLED OST_IMAGES_HE_INSTALLED OST_IMAGES_SSH_KEY; do
+for i in OST_IMAGES_BASE OST_IMAGES_NODE OST_IMAGES_ENGINE_INSTALLED OST_IMAGES_HOST_INSTALLED OST_IMAGES_HE_INSTALLED OST_IMAGES_STORAGE_BASE OST_IMAGES_SSH_KEY; do
   [[ -r "${!i}" ]] || declare $i=/usr/share/ost-images/$(basename ${!i})
   echo "${i} ${!i} $(rpm -qf ${!i} 2>/dev/null)"
+  pkglist="$(dirname ${!i})/$(basename ${!i} .qcow2)-pkglist.txt"
+  [[ -r "${pkglist}" ]] && cp "${pkglist}" "${OST_REPO_ROOT}/exported-artifacts/"
```
michalskrivanek, line 35: I don't think it should be done in ost_images. That's a helper that resolves image paths, it should not copy anything. I think handling this within lagofy at ost_init time using only the actually used images would be better. It can even be a part of root disk creation code perhaps, that's where you have that variable at hand, or just jq the $ost_conf for .vms[].root_disk_var

michalskrivanek (changes requested):

common/helpers/ost-images.sh
acernek: Done

lagofy.sh
```diff
@@ -240,6 +240,10 @@ ost_init() {
             qemu-img create -q -f qcow2 -b ${!vm_rootdisk_var} -F qcow2 $OST_ROOTDISK
             echo -n "root($(basename ${!vm_rootdisk_var})) "
 
+            # export package list
+            pkglist="$(dirname ${!vm_rootdisk_var})/$(basename ${!vm_rootdisk_var} .qcow2)-pkglist.txt"
+            [[ -r "${pkglist}" ]] && cp "${pkglist}" "${OST_REPO_ROOT}/exported-artifacts/"
```
tinez, line 245: Works fine, but can we put the package lists into a separate dir? i.e. `exported-artifacts/pkglist/`

lagofy.sh
```diff
@@ -170,6 +170,8 @@ ost_init() {
     mkdir "$OST_DEPLOYMENT/images"
     chcon -t svirt_image_t "$OST_DEPLOYMENT/images"
 
+    mkdir "${OST_REPO_ROOT}/exported-artifacts/package_lists"
```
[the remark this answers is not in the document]
michalskrivanek: we don't always clean exported-artifacts

michalskrivanek:
LGTM

michalskrivanek (approved):

PR #219, opened 2022-07-20
vm-template: Improve disk driver configuration

nirs:
- Use cache='none' to avoid double caches on host and gust. With
  cache='none' writes bypass the host page cache, and less memory is
  used on the host for caching. This also enabled usage of io='native'.

- Use io='native' which is the recommended configuration. With this I/O
  is submitted using single thread using libaio. This is usually faster
  than io='threads', and more important for OST, it uses one thread
  instead of multiple threads on qemu side, which keep the load on the
  host lower. This can improve reliability and performance when running
  many vms with few CPUs (as seem to happen in OST).

- Use discard='unmap' so discard requests punch holes in the images.
  I'm not sure it is useful for qcow2 images but it is a good default,
  in particular for developers running OST locally on small disks.

- Use iothread='1' - Enable I/O thread for the disk. We configured one
  I/O thread but it was not used. This can improve performance and
  reduce the chance of of blocking qemu when handling I/O on the event
  loop thread.

### DIFF
--- a/common/libvirt-templates/vm_template
+++ b/common/libvirt-templates/vm_template
@@ -38,7 +38,7 @@
   <devices>
     <emulator>/usr/libexec/qemu-kvm</emulator>
     <disk type='file' device='disk'>
-      <driver name='qemu' type='qcow2'/>
+      <driver name='qemu' type='qcow2' cache='none' io='native' discard='unmap' iothread='1'/>
       <source file='@OST_ROOTDISK@'/>
       <target dev='vda' bus='virtio'/>
       <boot order='1'/>


common/libvirt-templates/vm_template
```diff
@@ -42,7 +42,7 @@
       <source file='@OST_ROOTDISK@'/>
       <target dev='vda' bus='virtio'/>
       <boot order='1'/>
-      <serial>1</serial>
+      <serial>ost-root-disk</serial>
```
nirs, line 45: This breaks the tests since they expect to find disk `vda_1` and how the disk is named `ost_root_disk`:

```
<failure message="RuntimeError: Could not find block_vda_1 device in host devices: block_vda_ost_root_disk; computer; net_br_int_72_e5_b3_ad_7c_72; net_eth0_54_52_c0_a8_ca_03; net_eth1_54_52_c0_a8_c9_02; net_eth2_54_52_c0_a8_c9_03; net_eth3_54_52_c0_a8_c8_03; net_ip_vti0_00_00_00_00; net_lo_00_00_00_00_00_00; net_ovs_system_9e_ba_0a_ed_28_20; pci_0000_00_00_0; pci_0000_00_01_0; pci_0000_00_01_1; pci_0000_00_01_2; pci_0000_00_01_3; pci_0000_00_01_4; pci_0000_00_01_5; pci_0000_00_01_6; pci_0000_00_01_7; pci_0000_00_1f_0; pci_0000_00_1f_2; pci_0000_00_1f_3; pci_0000_01_00_0; pci_0000_02_00_0; pci_0000_03_00_0; pci_0000_04_00_0; pci_0000_05_00_0; pci_0000_06_00_0; pci_0000_07_00_0; scsi_host0; scsi_host1; scsi_host2; scsi_host3; scsi_host4; scsi_host5;">hosts_service = &lt;ovirtsdk4.services.HostsService object at 0x7fcd2d3c3820&gt;
```

I'll send this change in a separate pr since it is not related to the disk configuration change.

nirs: Change move to #220

aesteve-rh (approved):

aesteve-rh:
Small typo in the commit message: 'Use cache='none' to avoid double caches on host and gu**e**st.'. Other than that, LGTM.

PR #222, opened 2022-07-21
Improve disk_template driver configuration

nirs:
Use the same configuration for the disk_template as we use for the root
disk in vm_template. This is more important since the storage vm does
more I/O.

Fixes e2453ad536ef (vm-template: Improve disk driver configuration)

### DIFF
--- a/common/libvirt-templates/disk_template
+++ b/common/libvirt-templates/disk_template
@@ -1,5 +1,5 @@
 <disk type='file' device='disk'>
-  <driver name='qemu' type='raw' cache='writeback' discard='unmap'/>
+  <driver name='qemu' type='raw' cache='none' io='native' discard='unmap'/>
   <source file='@DISK_FILE@' index='2'/>
   <target dev='@DISK_DEV@' bus='scsi'/>
   <serial>@DISK_SERIAL@</serial>
--- a/common/libvirt-templates/vm_template
+++ b/common/libvirt-templates/vm_template
@@ -45,6 +45,9 @@
       <serial>ost-root-disk</serial>
     </disk>
     @DISKS@
+    <controller type="scsi" index="0" model="virtio-scsi">
+      <driver iothread="1"/>
+    </controller>
     <controller type='usb' model='none'/>
     <serial type='pty'>
       <source path='@SERIALLOG@'/>


michalskrivanek (approved):

aesteve-rh (approved):

common/libvirt-templates/disk_template
```diff
@@ -1,5 +1,5 @@
 <disk type='file' device='disk'>
-  <driver name='qemu' type='raw' cache='writeback' discard='unmap'/>
+  <driver name='qemu' type='raw' cache='none' io='native' discard='unmap' iothread='1'/>
   <source file='@DISK_FILE@' index='2'/>
   <target dev='@DISK_DEV@' bus='scsi'/>
```
nirs, line 4: It does not work since these disk use scsi instead of virtio. With scsi we need to set the iothread on the scsi controller. I don't see a reason to use scsi - we get better
performance with virtio, but we can change this later.

michalskrivanek: I don't quite recall...it could be we use scsi due to discard. but i don't really know.
feel free to add a controller, it may be way easier than dealing with the fallout of sdX vs vdX

nirs: Added a controller, converting to virtio can be done later. There is no reason to use scsi for discard, virtio works fine.

nirs:
Disk xml with this change:

```
    <disk type='file' device='disk'>
      <driver name='qemu' type='raw' cache='none' io='native' discard='unmap'/>
      <source file='/home/jenkins/workspace/ds-ost-baremetal_manual/deployment/images/storage-sda.qcow2' index='2'/>
      <backingStore/>
      <target dev='sda' bus='scsi'/>
      <serial>2</serial>
      <alias name='scsi0-0-0-0'/>
      <address type='drive' controller='0' bus='0' target='0' unit='0'/>
    </disk>
    <disk type='file' device='disk'>
      <driver name='qemu' type='raw' cache='none' io='native' discard='unmap'/>
      <source file='/home/jenkins/workspace/ds-ost-baremetal_manual/deployment/images/storage-sdb.qcow2' index='1'/>
      <backingStore/>
      <target dev='sdb' bus='scsi'/>
      <serial>3</serial>
      <alias name='scsi0-0-0-1'/>
      <address type='drive' controller='0' bus='0' target='0' unit='1'/>
    </disk>
    <controller type='scsi' index='0' model='virtio-scsi'>
      <driver iothread='1'/>
      <alias name='scsi0'/>
      <address type='pci' domain='0x0000' bus='0x03' slot='0x00' function='0x0'/>
    </controller>
```

PR #225, opened 2022-07-25
engine: Use drop-in for logging configuration

tinez:
Requires: https://github.com/oVirt/ovirt-engine/pull/553

### DIFF
--- a/ost_utils/ansible/collection.py
+++ b/ost_utils/ansible/collection.py
@@ -79,9 +79,21 @@ def engine_setup(
     host_name = socket.gethostname()
     host_ip = socket.gethostbyname(host_name)
 
+    config = f'''\
+SSO_ALTERNATE_ENGINE_FQDNS="${{SSO_ALTERNATE_ENGINE_FQDNS}} {host_ip} {host_name} {engine_ip}"
+'''
+
+    if os.environ.get('ENABLE_DEBUG_LOGGING'):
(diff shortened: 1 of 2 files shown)


michalskrivanek (changes requested):
Please keep it conditional, I like to run with default logging. 

I don't think we needed multiple vars, but ok...

ost_utils/ansible/collection.py
```diff
@@ -79,6 +79,12 @@ def engine_setup(
     host_name = socket.gethostname()
     host_ip = socket.gethostbyname(host_name)
 
+    ansible_engine.copy(
+        content='ORG_OVIRT_LOG_LEVEL="DEBUG"\\nKEYCLOAK_LOG_LEVEL="DEBUG"\\nCORE_BLL_LOG_LEVEL="DEBUG"\\nROOT_LOG_LEVEL="DEBUG"',  # noqa: E501
+        dest='/etc/ovirt-engine/engine.conf.d/98-logging.conf',
```
michalskrivanek, line 84: also, let's unite the two confs. ansible is horrible, not adding a new call is a plus.

tinez: Ack

tinez:
> Please keep it conditional, I like to run with default logging.

Done.

michalskrivanek (approved):
thanks

PR #229, opened 2022-07-28
ost.sh: Add copy command

acernek:


michalskrivanek:
I would find it convenient to configure /etc/ssh/ssh_config with StrictHostKeyChecking=no, UserKnownHostsFile=/dev/null. Nothing more.
Then it's easy to use for any ssh/scp operation

tinez:
> I would find it convenient to configure /etc/ssh/ssh_config with StrictHostKeyChecking=no, UserKnownHostsFile=/dev/null. Nothing more. Then it's easy to use for any ssh/scp operation

I'd rather vote for a drop-in config instead in `/etc/ssh/ssh_config.d` - one that would set those vars for hosts matching `ost-*`.
Anyway, I think this should be done in a separate PR.

tinez:
I'm having hard time trying to figure out how the new command works. @acernek could you please add specific examples to the help message?

michalskrivanek:
> > I would find it convenient to configure /etc/ssh/ssh_config with StrictHostKeyChecking=no, UserKnownHostsFile=/dev/null. Nothing more. Then it's easy to use for any ssh/scp operation
> 
> I'd rather vote for a drop-in config instead in `/etc/ssh/ssh_config.d` - one that would set those vars for hosts matching `ost-*`. 

of course! that's what i meant but didn't write:)

> Anyway, I think this should be done in a separate PR.

but i dont' see a point of this one then. special operation for something so simple as cp is IMO just adds bloat.

tinez:
> but i dont' see a point of this one then. special operation for something so simple as cp IMO just adds bloat.

I still find it useful - using regular scp requires you to:
1. find out the ip address of the VM
2. pass all those annoying options on the commandline

We have the `./ost.sh shell` convenience wrapper and `./ost.sh copy` would be the symmetric `scp` equivalent.

tinez (approved):

PR #232, opened 2022-08-01
Revert "basic: el8stream: Downgrade python on host-0 temporarily"

tinez:
Reverts oVirt/ovirt-system-tests#231


nirs:
Do we have results from the weekend? If this eliminates the stuck vdsm issue,
it is better to run it for a week to be sure.

tinez:
Unfortunately this didn't solve the issue - we still observe vdsm being stuck from time to time. We need to approach this problem from a different angle.

PR #233, opened 2022-08-02
deployment: Wait for 'default.target'

tinez:
Waiting for 'sshd' to be responsive is sometimes not enough.
We have cases of OST runs failing where 'sshd' is already responsive,
but 'NetworkManager' is still doing some configuration.

Change-Id: I5542971d6aef1f0a8eb2ec3d2acd006649c95570

### DIFF
--- a/ost_utils/pytest/fixtures/deployment.py
+++ b/ost_utils/pytest/fixtures/deployment.py
@@ -14,9 +14,11 @@ import pprint
 
 import pytest
 
+from ost_utils import assert_utils
 from ost_utils import coverage
 from ost_utils import deployment_utils
 from ost_utils import utils
+from ost_utils.ansible import AnsibleExecutionError
 from ost_utils.deployment_utils import package_mgmt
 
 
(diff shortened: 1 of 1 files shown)


ost_utils/pytest/fixtures/deployment.py
```diff
@@ -112,8 +114,10 @@ def deploy(
         LOGGER.info("Environment already deployed")
         return
 
-    LOGGER.info("Waiting for SSH on the VMs")
-    ansible_all.wait_for_connection(timeout=120)
+    LOGGER.info("Waiting for VMs to fully boot")
+    assert_utils.true_within_short(
+        ansible_all.shell("systemctl is-active default.target"), allowed_exceptions=[AnsibleExecutionError]
```
acernek, line 119: you have to pass a function (i.e. lamda)

tinez: right, wrong copy-paste :)

acernek (approved):
LGTM, assuming assert was left out on purpose (previous code was not asserted either)

michalskrivanek:
needs also a feedback when it succeeds

tinez:
It's still wrong, will update in a sec

michalskrivanek:
> It's still wrong, will update in a sec

yeah, just noticed too. nevermind about the extra logs/feedback, i'll post something on top, it's bothering me long enough...

michalskrivanek (approved):
still using ansible, meh, but it hopefully works:)

michalskrivanek:
it works for me manually, we will only get feedback when we use this massively, let's go...

PR #237, opened 2022-08-10
UI tests: Add tests for edit and details views

ljelinkova:


ost_utils/selenium/page_objects/EntityDialog.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def is_displayed(self):
+        modal_title = self.ovirt_driver.driver.find_element(By.CLASS_NAME, 'modal-title')
+        spinner_displayed = self.ovirt_driver.is_class_name_present('spinner')
+        return modal_title.text == f'{self.action} {self.entity_type}' and not spinner_displayed
+
+    def get_displayable_name(self):
+        return f'{self.action} {self.entity_type} dialog'
+
+    def ok(self, handle_warning_dialogs=lambda *args: None):
```
tinez, line 32: The `handle_warning_dialogs` argument is horrible :D since the `page_objects` module generally sticks with OOP I'd rather have an empty `EntityDialog.handle_ok_popups` method that can be overridden in subclasses

ost_utils/selenium/page_objects/HostListView.py
```diff
@@ -14,6 +16,20 @@ def __init__(self, ovirt_driver):
...
+        detail_view = HostDetailView(self.ovirt_driver, self.breadcrumbs, name)
+        detail_view.wait_for_displayed()
+        return detail_view
+
+    def edit(self, name):
+        super().edit(name)
+
+        dialog = HostDialog(self.ovirt_driver, 'Edit')
+        dialog.wait_for_displayed()
+        return dialog
```
tinez, line 31: I gotta admit - I'm not a big fan of OOP - why does the base class implementation has a different return type then the subclass override? I won't get grumpy with this one - we'd need a bigger refactor to handle it.

ost_utils/selenium/page_objects/VmDialog.py
```diff
@@ -0,0 +1,20 @@
...
+from .EntityDialog import EntityDialog
+
+
+class VmDialog(EntityDialog):
+    def __init__(self, ovirt_driver, action):
+        super(VmDialog, self).__init__(ovirt_driver, 'Virtual Machine', action)
+
+    def is_displayed(self):
+        cluster_element_text = self.ovirt_driver.driver.find_element(By.ID, 'VmPopupWidget_dataCenterWithCluster').text
+        return super().is_displayed() and cluster_element_text
```
tinez, line 15: Is the `and cluster_element_text` part enough? Don't we need to compare to some string value?

basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
```diff
@@ -315,6 +359,28 @@ def test_templates(ovirt_driver, cirros_image_template_name):
...
+
+    # Test the edited values in the details view
+    template_detail_view = template_list_view.open_detail_view(blank_template_name)
+    assert template_detail_view.get_name() == blank_template_name
+    assert template_detail_view.get_description() == blank_template_description
+    save_screenshot('blank-template-detail')
+
+    webadmin_menu.open_template_list_view()
+    template_detail_view = template_list_view.open_detail_view(imported_template)
+    assert template_detail_view.get_name() == imported_template
```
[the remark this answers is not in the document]
ljelinkova: Not really, if the assertions fail the screenshot will be taken after the test method finishes.

ost_utils/selenium/page_objects/VmDialog.py
ljelinkova: yes, it is enough - if the `cluster_element_text ` is non empty text, the dialog is loaded.

michalskrivanek:
lgtm

ost_utils/selenium/page_objects/HostListView.py
tinez: > Would it be possible to pass the `HostDialog` from `HostListView` to `EntityListView`? E.g. in the constructor?

That's maybe a bit too complicated for our use case. Here I'd suggest making `EntityListView` an abstract base class. That still allows you to have concrete methods implemented. Here's a simple example:

```
import abc

class EntityListView(abc.ABC):
    def _click_edit(self):
        print("hi zorro")

    @abc.abstractmethod
    def open_edit_dialog(self):
        pass

class HostDialog(EntityListView):
    def open_edit_dialog(self):
        self._click_edit()
        ...
```

This code won't allow instantiating classes derived from `EntityListView` that don't have `open_edit_dialog` implemented

tinez: Oh, and I wouldn't squeeze that change into this PR - it's big enough already.

tinez (approved):

PR #238, opened 2022-08-11
UI tests: Improve logout

ljelinkova:
Sometimes when logging out, instead of displaying the welcome screen, the page stays on dashboard. Probable cause is that some queries are still running and interrupt the page redirect.

This page manually refreshes the webadmin so that the welcome screen is displayed.

### DIFF
--- a/ost_utils/selenium/page_objects/WebAdminTopMenu.py
+++ b/ost_utils/selenium/page_objects/WebAdminTopMenu.py
@@ -4,8 +4,8 @@
 #
 import logging
 
-from selenium.webdriver.common.by import By
 from .Displayable import Displayable
+from .WelcomeScreen import WelcomeScreen
 
 LOGGER = logging.getLogger(__name__)
 
@@ -15,7 +15,7 @@ class WebAdminTopMenu(Displayable):
         super(WebAdminTopMenu, self).__init__(ovirt_driver)
 
     def is_displayed(self):
-        return self.ovirt_driver.find_element(By.TAG_NAME, 'nav').is_displayed()
+        return self.ovirt_driver.is_css_selector_displayed('nav')
 
     def get_displayable_name(self):
         return 'WebAdmin top menu'
@@ -30,3 +30,21 @@ class WebAdminTopMenu(Displayable):
         )
         self.ovirt_driver.xpath_wait_and_click('User dropdown menu', '//*[@id="HeaderView_userName"]')
         self.ovirt_driver.xpath_wait_and_click('Logout menu', '//*[@id="HeaderView_logoutLink"]')
+
+        self.ovirt_driver.wait_until(
+            'The welcome screen is not displayed after logout', self._welcome_screen_displayed
+        )
+
+    def _welcome_screen_displayed(self):
+        top_menu_displayed = self.is_displayed()
+        welcome_screen_displayed = WelcomeScreen(self.ovirt_driver).is_displayed()
+
+        if welcome_screen_displayed:
+            return True
+        elif not top_menu_displayed:
+            # Page is not fully loaded yet
+            return False
+        else:
+            LOGGER.debug('The top menu is still displayed, refreshing the page')
+            self.ovirt_driver.refresh()
+            return False
--- a/ost_utils/selenium/navigation/driver.py
+++ b/ost_utils/selenium/navigation/driver.py
@@ -30,6 +30,9 @@ class Driver:
     def get(self, url):
         self.__driver.get(url)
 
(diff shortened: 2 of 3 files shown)


tinez (commented):
Hmm can we simplify the big picture approach? Not sure if I understand the algorithm correctly, but maybe we can have a single `wait_until` that would check for `not top_menu_is_displayed and welcome_screen_displayed`?

ost_utils/selenium/page_objects/WebAdminTopMenu.py
```diff
@@ -30,3 +31,35 @@ def logout(self):
...
         self.ovirt_driver.xpath_wait_and_click('User dropdown menu', '//*[@id="HeaderView_userName"]')
         self.ovirt_driver.xpath_wait_and_click('Logout menu', '//*[@id="HeaderView_logoutLink"]')
+
+        self.ovirt_driver.wait_while(
+            'The admin portal is still displayed after logout and refresh', self._is_displayed_with_refresh
+        )
+
+    def _is_displayed_with_refresh(self):
+        if self._is_displayed_with_error_handling(self.is_displayed):
+            print('Still displayed')
```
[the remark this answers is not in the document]
ljelinkova: The big picture is that sometimes we try to log out and the user actually is logged out but while redirecting to welcome screen some other queries jump in and it all ends up by staying on the dashboard page. In reality, as soon as you'd click on anything or reload the page the system would realized that the user is not logged in and it would redirect you to the welcome page (so not really a big security issue worth investing the time to investigate). 

The reload of the page is important  and we cannot just wait for `not top_menu_is_displayed and welcome_screen_displayed`. To be honest, there is a lot of guessing in this patch. If you prefer, I can add some debug logs and we can monitor the tests to see if it actually needs to be refreshed repeatedly but given that the issue is rare I am not sure it is worth it. If the issue occurs I am OK with refreshing it more times if it makes tests pass.

Sorry about the `print`, a shameful leftover :-)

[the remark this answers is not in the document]
tinez: So maybe I wasn't clear enough - I don't want to get rid of the refresh. This is more or less what I had in mind:
```
wait_until(welcome_screen_displayed)

def welcome_screen_displayed():
    top_menu_displayed = ...
    welcome_screen_displayed = ...
    if top_menu_displayed or not welcome_screen_displayed:
        refresh()
        return False
    return True
```

michalskrivanek (approved):

ost_utils/selenium/page_objects/WebAdminTopMenu.py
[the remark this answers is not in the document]
ljelinkova: I see, I've changed the code according your suggestion (with some necessary tweaks)

ost_utils/selenium/page_objects/WebAdminTopMenu.py
```diff
@@ -30,3 +31,29 @@ def logout(self):
...
+    def _welcome_screen_displayed(self):
+        top_menu_displayed = self._is_displayed_with_error_handling(self.is_displayed)
+        welcome_screen_displayed = self._is_displayed_with_error_handling(
+            WelcomeScreen(self.ovirt_driver).is_displayed
+        )
+
+        if welcome_screen_displayed:
+            return True
+        elif not top_menu_displayed:
+            # None page is fully loaded yet
```
michalskrivanek, line 48: what does it mean?

ljelinkova: That after we reload the page, there is neither top menu nor welcome screen (yet). We should wait a bit longer to see which one will be displayed.

ost_utils/selenium/page_objects/WebAdminTopMenu.py
```diff
@@ -30,3 +31,29 @@ def logout(self):
...
+            return False
+        else:
+            LOGGER.debug('The top menu is still displayed, refreshing the page')
+            self.ovirt_driver.driver.refresh()
+            return False
+
+    def _is_displayed_with_error_handling(self, is_displayed):
+        try:
+            return self.ovirt_driver.retry_if_stale(is_displayed)
+        except Exception:
```
tinez, line 58: Here comes my usual complaint about using `Exception` :) Can we narrow down the exception type here?

ost_utils/selenium/page_objects/WebAdminTopMenu.py
tinez: I think the current comment is good enough

PR #240, opened 2022-08-16
Hide driver

ljelinkova:
UI tests: do not allow direct access to selenium webdriver

This patch makes the selenium driver instance private so that the clients are forced to use ovirt driver. This way we can better handle specific errors. The patch enhances the find_element(s) method with retry for known issues


michalskrivanek:
I think a "KnownIssue" is not very helpful as we use it for these retries of mysterious unknown exceptions/timeouts in selenium. So quite the opposite of "known":) but ok...

PR #244, opened 2022-08-22
UI tests: Fix login screen after logout

ljelinkova:
Sometimes, the login screen is displayed after the logout. In those case, navigate directly to the welcome screen so that we end up consistently on the welcome screen after logout.


ljelinkova:
It seems that https://github.com/oVirt/ovirt-system-tests/pull/246 fixed the issue, so this PR is not needed any more.

PR #247, opened 2022-08-24
UI tests: add test for cluster upgrade

ljelinkova:


basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
```diff
@@ -281,6 +283,41 @@ def test_clusters(ovirt_driver, save_screenshot, selenium_browser_name, ost_clus
...
+    upgrade_dialog.next()
+
+    save_screenshot('cluster-upgrade-dialog-review')
+    upgrade_dialog.upgrade()
+
+    save_screenshot('cluster-upgrade-dialog-progress')
+
+    events_view = upgrade_dialog.go_to_event_log()
+    assert assert_utils.true_within_short(lambda: events_view.events_contain('Cluster upgrade progress: 0%'))
+    assert assert_utils.true_within_short(lambda: events_view.events_contain('Successfully set upgrade running flag'))
```
michalskrivanek, line 309: can you please check cluster policy change too?

michalskrivanek:
we can do a followup, but i would really like to see a real package being updated (i.e. downgrade something, run cluster upgrade, check it's updated e.g. on one host and not on the other if you select just one in dialog). we can add something to the dummy repo. @tinez what do you think?

basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
ljelinkova: I could add a REST API call to check that, but I would prefer if the change was also mentioned in the events (it is not at the moment). Any idea if we can get it there? These are UI tests and we should be able to assert things in the UI only (yeah, I guess we could open the edit dialog, but having an event would be nicer)

ljelinkova:
OST failed on an unrelated change (logout from user portal)

basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
```diff
@@ -281,6 +283,41 @@ def test_clusters(ovirt_driver, save_screenshot, selenium_browser_name, ost_clus
...
+    save_screenshot('cluster-upgrade-dialog-options')
+    upgrade_dialog.next()
+
+    save_screenshot('cluster-upgrade-dialog-review')
+    upgrade_dialog.upgrade()
+
+    save_screenshot('cluster-upgrade-dialog-progress')
+
+    events_view = upgrade_dialog.go_to_event_log()
+    assert assert_utils.true_within_short(lambda: events_view.events_contain('Cluster upgrade progress: 0%'))
```
tinez, line 308: Maybe instead of copying the logic for all the events it would be nicer to have a list of expected events and a loop, i.e.:
```python
expected_events = ['Cluster upgrade..', 'Successfully set upgrade...', ...]
for ee in expected_events:
    assert assert_utils.true_within...
```

ost_utils/selenium/page_objects/EventsView.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class EventsView(Displayable, WithBreadcrumbs):
+    def __init__(self, ovirt_driver):
+        super(EventsView, self).__init__(ovirt_driver)
+
+    def is_displayed(self):
+        current_breadcrumbs = self.get_breadcrumbs()
+        breadcrumbs_present = (
+            len(current_breadcrumbs) == 2 and current_breadcrumbs[0] == 'Events' and current_breadcrumbs[1] == 'Events'
```
tinez, line 21: It's simpler to compare two lists directly instead. Not sure what type `current_breadcrumbs` is, but this should work:
```python
list(current_breadcrumbs) == ['Events', 'Events']
```

ost_utils/selenium/page_objects/EventsView.py
```diff
@@ -0,0 +1,45 @@
...
+    def events_contain(self, event_substring):
+        return any(event_substring in event for event in self.get_events())
+
+    def _get_events(self):
+        events_entities = self.ovirt_driver.find_elements(
+            By.XPATH,
+            '//div[contains(@id, "col2")]',
+        )
+        events = []
+        for events_entity in events_entities:
```
tinez, line 42: Let's use list comprehension instead.

tinez:
> we can do a followup, but i would really like to see a real package being updated (i.e. downgrade something, run cluster upgrade, check it's updated e.g. on one host and not on the other if you select just one in dialog). we can add something to the dummy repo. @tinez what do you think?

Well, following our "keep it offline" philosophy we should cache the down-/upgraded packages during ost-images time. The repo holding the package should be local too. Are you thinking about some dummy package or a real one?

basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
ljelinkova: I prefer it this way - the events are not there at once, the asserts wait for them to gradually appear. This allows us to add some extra checks using REST API if needed (as Michal suggested to check if the cluster's policy has changed) + every assert uses just short timeout, whereas we would need to wait the long timeout for all of them to finish to just see that the process has not started at all.

tinez: Fair enough :)

tinez (approved):

PR #248, opened 2022-08-24
UI tests: wait for the list view after closing the dialog

ljelinkova:
Sometimes, when closing the dialog, the validations might take a while. We should therefore wait for the list views to be displayed before continuing further with the tests.


michalskrivanek:
the commit message should probably explain why

ljelinkova:
Updated commit message and the PR's description

tinez (approved):

PR #252, opened 2022-08-30
UI tests: wrap more webdriver functions to retry_if_known_issue

ljelinkova:


tinez:
I admit - I don't precisely track all the selenium failures and I'm not exactly sure which calls are prone to those timeout exceptions, but this is a lot of stuff wrapped. Are we sure we want all of this in retries?

ljelinkova:
I think we do. It takes much more time to investigate failed job then to retry a method while the job is running. Let's admin, Selenium is not super reliable and wrapping it helps a lot.

PR #254, opened 2022-08-31
Take master SD type from env var

acernek:
So that we can easily switch the type.

### DIFF
--- a/ost_utils/pytest/fixtures/storage.py
+++ b/ost_utils/pytest/fixtures/storage.py
@@ -7,6 +7,7 @@
 import functools
 import pytest
 
+from ost_utils import constants
 from ost_utils import network_utils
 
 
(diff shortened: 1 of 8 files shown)


ost_utils/pytest/fixtures/storage.py
```diff
@@ -34,3 +36,18 @@ def sd_nfs_host_storage_name():
...
+
+
+@pytest.fixture(scope="session")
+def secondary_storage_domain_name(master_storage_domain_type):
+    return constants.SD_ISCSI_NAME if master_storage_domain_type == "nfs" else constants.SD_NFS_NAME
+
+
+@pytest.fixture(scope="session")
+def master_storage_domain_type():
+    return os.environ.get("MASTER_SD_TYPE", "nfs")
```
tinez, line 53: We keep the things that come from the env in `fixtures/env.py`, so let's move this fixture there. The names for the env vars are also prefixed with `OST_`, so let's do the same here.

tinez (changes requested):
Let's also use the new constant in `test_storage_domains` from UI tests.

ost_utils/pytest/fixtures/storage.py
acernek: Sorry, missed that part. Done

tinez (approved):

PR #261, opened 2022-09-13
Revert "Test that notifier works after restart"

tinez:
Reverts oVirt/ovirt-system-tests#198. This PR broke `test_verify_notifier` on rhel8. Needs further investigation.


mwperina:
It works fine on upstream. Can't we just disable for downstream tests until fixed instead of complete revert?

tinez:
> It works fine on upstream. Can't we just disable for downstream tests until fixed instead of complete revert?

The problem is these are not the newly added tests that are failing - it's the old one. I have no idea how adding that new setting could break this. Theoretically we could xfail all three notifier tests on rhel8, but given the weird behavior I figured it's maybe better for Eli to have a look at this.

PR #263, opened 2022-09-15
ansible: Fix new ansible-lint warnings

acernek:


tinez (changes requested):
Didn't pass CI

acernek:
> Didn't pass CI

Apparently we don't update the packages used for linting - doing that would however prolong the runs and there seem to be bug in the latest version of ansible-lint anyway, so I've partially reversed the change.

tinez (approved):

PR #265, opened 2022-09-16
UI tests: fix problem with logout from webadmin

ljelinkova:
When using Keycloak, the tests sometimes fail because the user is not logged out properly and the admin
stays logged in.


ljelinkova:
Lets try deleting cookies...

PR #271, opened 2022-09-21
UI tests: Do not logout user from webadmin when using Keycloak

ljelinkova:
The logout from Keycloak does not work properly and as a result, the admin user stayes logged in from time to time. This causes the tests to fail. As this is not likely to change in the near future, this patch leaves out logging from webadmin if the Keycloak is used. As a consequence, there is no login to the user portal required and the number of VMs is not checked (as that is different for the test user and the admin user).


michalskrivanek:
I think https://github.com/oVirt/ovirt-engine/pull/679/ actually solves the problem

PR #273, opened 2022-09-22
lagofy: use el9stream-engine-installed

michalskrivanek:
now that we build el9stream engine and have the corresponding image, we can
run el9stream runs with it


michalskrivanek:
el9 engine builds enabled by https://github.com/oVirt/ovirt-engine/pull/647

PR #274, opened 2022-09-22
basic: Check TPM presence

mz-pdm:
We enable TPM for vm0 and then check that a TPM device is actually present in the guest OS.

vm0 configuration must be changed to a TPM-compatible guest OS in order to be able to have a TPM device.


mz-pdm:
Hm, another platform bug fix not present in Stream 8. So we can have this test only after switching away from Stream 8.

mz-pdm:
There seems to be no good VM to run the test on:

- vm0 tests memory snapshots, which are incompatible with TPM.
- The other VMs don't have a bootable guest OS.

I don't think this tests is worth of larger changes, so closing.

PR #277, opened 2022-09-26
selenium fix

michalskrivanek:
- selenium: upgrade to 4.4.0
- selenium: export novnc port
- selenium: use Enter instead of clicking OK

### DIFF
--- a/ost_utils/pytest/fixtures/selenium.py
+++ b/ost_utils/pytest/fixtures/selenium.py
@@ -116,12 +116,14 @@ def selenium_browser(
     container_id = ansible_storage.shell(
         "podman run -d"
         f" -p {selenium_port}:{selenium_port}"
+        "  -p 7900:7900"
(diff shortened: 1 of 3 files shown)


ost_utils/pytest/fixtures/selenium.py
```diff
@@ -116,12 +116,14 @@ def selenium_browser(
     container_id = ansible_storage.shell(
         "podman run -d"
         f" -p {selenium_port}:{selenium_port}"
+        "  -p 7900:7900"
```
tinez, line 119: Why do we expose this port?

michalskrivanek: from commit message:
for debugging we want to be able to get to port 7900 where selenium
starts a passwordless no-vnc instance

tinez: ack

tinez (approved):

PR #278, opened 2022-09-27
UI tests: Workaround problem with keycloak logout

ljelinkova:
If the admin is not logged out properly, let's try to logout admin from the user portal and login test user.


ljelinkova:
Closing in favor of creating a completely new session for user portal and grafana

PR #280, opened 2022-09-28
ui: User portal and Grafana tests in a separate Selenium session

tinez:
- selenium: Remove unused 'selenium_version' fixture
- selenium: Move fixtures to fixtures.selenium for reuse
- ui: Move some fixtures to 'fixtures.ui' module
- ui: Remove unused imports and objects
- ui: Separate user portal and grafana tests from other UI tests


tinez:
Right, so in the latest run `test_logout` failed and it's not really relevant with the current PR, so we can remove it.

basic-suite-master/test-scenarios/test_100_ui_webadmin.py
```diff
@@ -688,82 +538,8 @@ def test_logout(ovirt_driver, engine_webadmin_url, keycloak_enabled):
     # navigate directly to welcome page to prevent problems with redirecting to login page instead of welcome page
     ovirt_driver.get(engine_webadmin_url)
 
-    welcome_screen = WelcomeScreen(ovirt_driver)
-    welcome_screen.wait_for_displayed()
-    welcome_screen.wait_for_user_logged_out()
-    assert welcome_screen.is_user_logged_out()
-
-    if keycloak_enabled:
```
ljelinkova, line 696: I think we should still keep this test and assert for non-keycloak authentication.

tinez: I think I actually implemented it the way you suggested - it's still running for non-keycloak envs. The diff is big so please have a second look.

michalskrivanek:
I don't know...so what does this actually solve and how?

tinez:
> I don't know...so what does this actually solve and how?

It solves the problem of user portal tests failing. We're simply running a completely new Selenium setup for user portal and Grafana tests. The gist of the PR is changing `scope=session` to `scope=module`. The rest is just shuffling stuff around and a cleanup.

michalskrivanek:
and what is it going to help with? IIUC it won't necessarily affect the logout/keycloak problem

basic-suite-master/test-scenarios/test_100_ui_webadmin.py
ljelinkova: yes, I can see it now, sorry about that.

ljelinkova (approved):

PR #281, opened 2022-09-29
selenium: Go back to 4.0.0

tinez:
The 4.4.0 Selenium version is more unstable than 4.0.0.


ljelinkova (approved):

michalskrivanek:
is 4.0.0 really that much better? eh...we want modern browsers too...

sjd78:
Would the `requirements.txt` selenium version also need to be changed to match?

michalskrivanek:
we'll go with 4.5.0

PR #282, opened 2022-09-29
clean sessions

michalskrivanek:
- selenium: export novnc port
- selenium: use Enter instead of clicking OK
- try to clean up any SSO sessions before entering UI tests


michalskrivanek:
nope, there still is an api token allocated first and the ui session doesn't get a refresh token then

PR #283, opened 2022-09-30
Use `shadow_root` for ui-extensions modals

sjd78:
To accommodate PR https://github.com/oVirt/ovirt-engine-ui-extensions/pull/73 that uses shadow DOM to render modals from ovirt-engine-ui-extensions, adjust the selenium tests to use the `shadow_root` when necessary.  Querying elements from the base document do not recurse into shadow-roots.  See https://github.com/oVirt/ovirt-engine-ui-extensions/pull/73#issuecomment-1259760170 for more explanation.

Some more info on shadow dom in selenium: https://titusfortner.com/2021/11/22/shadow-dom-selenium.html

The changes will be backwards compatible with non shadow dom modals.

Change details:

`driver`
  - add type info so Intellisense works within the class
  - add `find_dialog_root`, `find_dialog_element` and related functions to more easily handle testing of dialogs/modals/popups
  - enhance `find_element` and `find_elements` functions to be usable against the shadow dom ui-extensions modals
  - maintain backward compatibility for old ui-extensions dialogs
  - Chrome and Firefox currently have different handling requirements for shadow DOM components.  The difference is accounted for in function `_access_shadow_root`.

`ClusterListView`
  - split out ClusterUpgradeDialog to a seperate file

`ClusterUpgradeDialog`
  - use the ui-extensions enhanced `find_element` and `find_elements` functions to handle both shadowRoot and normal DOM structures

`VmVgpuDialog`
  - use the ui-extensions enhanced `find_element` and `find_elements` functions to handle both shadowRoot and normal DOM structures

`WithOvirtDriver`
  - add type info so Intellisense works in subclasses

### DIFF
--- a/ost_utils/selenium/navigation/driver.py
+++ b/ost_utils/selenium/navigation/driver.py
@@ -101,23 +101,21 @@ class Driver:
                 "return arguments[0].shadowRoot.querySelector('.ui-extensions-plugin-root')", container
             )
 
-    def _find_dialog_root_now(self, modal_id: str, driver: WebDriver = None) -> WebElement:
-        _driver = driver if driver else self.__driver
+    def _find_dialog_root_now(self, modal_id: str) -> WebElement:
         try:
             # prefer the shadowDom version
-            shadow_host = _driver.find_element(By.ID, f'shadow-root-container-{modal_id}')
+            shadow_host = self.__driver.find_element(By.ID, f'shadow-root-container-{modal_id}')
             return self._access_shadow_root(shadow_host)
         except NoSuchElementException:
             # if not found, look for a normal modal
-            old_modal = _driver.find_element(By.ID, modal_id)
-            return old_modal
+            return self.__driver.find_element(By.ID, modal_id)
 
     def _find_dialog_root(self, modal_id: str, immediate: bool = True) -> WebElement:
         if immediate:
             return self._find_dialog_root_now(modal_id)
         else:
             dialog_root = WebDriverWait(self.__driver, timeout=assert_utils.SHORT_TIMEOUT).until(
-                lambda driver: self._find_dialog_root_now(modal_id, driver)
+                lambda driver: self._find_dialog_root_now(modal_id)
             )
             return dialog_root
 
--- a/ost_utils/selenium/navigation/driver.py
+++ b/ost_utils/selenium/navigation/driver.py
@@ -13,6 +13,8 @@ from selenium.common.exceptions import (
 )
 from selenium.webdriver.common.action_chains import ActionChains
 from selenium.webdriver.common.by import By
+from selenium.webdriver.remote.webdriver import WebDriver
+from selenium.webdriver.remote.webelement import WebElement
 from selenium.webdriver.support.ui import WebDriverWait
 from urllib3.exceptions import ReadTimeoutError
 
@@ -22,7 +24,7 @@ LOGGER = logging.getLogger(__name__)
 
 
 class Driver:
-    def __init__(self, driver):
+    def __init__(self, driver: WebDriver):
 
         # this is a selenium webdriver instance, keeping it private so
         # that we can wrap and control it better (e.g. retry after some exceptions)
@@ -69,6 +71,52 @@ class Driver:
             # If failed, lets try again with waiting
             LOGGER.exception(f'!!! find_elements() failed with {e.__class__.__name__}, retrying')
             return self.retry_if_known_issue(self.__driver.find_elements, by, value)
+
+    def _access_shadow_root(self, container: WebElement) -> WebElement:
+        try:
+            return container.shadow_root
+        except AssertionError:
+            # If the WebDriver's browser in use does NOT support `shadow_root`, then we need
+            # to fallback to using JavaScript.  The div containing the component will provide
+            # a valid WebElement that can be used just like the normal `shadow_root`.
+            return self.__driver.execute_script(
+                "return arguments[0].shadowRoot.querySelector('.ui-extensions-plugin-root')", container
+            )
+
+    def _find_dialog_root(self, modal_id: str, driver: WebDriver = None) -> WebElement:
+        _driver = driver if driver else self.__driver
+        try:
+            # prefer the shadowDom version
+            shadow_host = _driver.find_element(By.ID, f'shadow-root-container-{modal_id}')
+            return self._access_shadow_root(shadow_host)
+        except NoSuchElementException:
+            # if not found, look for a normal modal
+            old_modal = _driver.find_element(By.ID, modal_id)
+            return old_modal
+
+    def find_dialog_root(self, modal_id: str, immediate: bool = False) -> WebElement:
+        if immediate:
+            return self._find_dialog_root(modal_id)
+        else:
+            dialog_root = WebDriverWait(self.__driver, timeout=5).until(lambda d: self._find_dialog_root(modal_id, d))
+            return dialog_root
+
+    def find_dialog_element(self, modal_id: str, locator, waitCondition=None, message=None) -> WebElement:
+        if waitCondition is None:
+            dialog_root = self.find_dialog_root(modal_id)
+            return dialog_root.find_element(*locator)
+        else:
+
+            def __finder(driver: WebDriver):
+                dialog_root = self._find_dialog_root(modal_id, driver)
+                el = dialog_root.find_element(*locator)
+                if el and waitCondition(el):
+                    return el
+                else:
+                    return False
+
+            el = WebDriverWait(self.__driver, timeout=5).until(__finder, message)
+            return el
 
     def execute_script(self, script):
         self.retry_if_known_issue(self.__driver.execute_script, script)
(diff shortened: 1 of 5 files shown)


sjd78:
Note: The VM vGPU tests (in class `VmVgpuDialog`) still need to be fixed.

sjd78:
Note: Changed to draft until I can verify it works with https://github.com/oVirt/ovirt-engine-ui-extensions/pull/73 applied.

sjd78:
ui-extension modal handling is working for old and new shadow-dom versions in both chrome and firefox

:smile:

sjd78:
Bump for review? @michalskrivanek, @leistnerova, @tinez

sjd78:
Bump for review @ljelinkova

ost_utils/selenium/navigation/driver.py
```diff
@@ -69,6 +71,52 @@ def find_elements(self, by, value):
             LOGGER.exception(f'!!! find_elements() failed with {e.__class__.__name__}, retrying')
             return self.retry_if_known_issue(self.__driver.find_elements, by, value)
 
+    def _access_shadow_root(self, container: WebElement) -> WebElement:
+        try:
+            return container.shadow_root
+        except AssertionError:
+            # If the WebDriver's browser in use does NOT support `shadow_root`, then we need
```
ljelinkova, line 78: Is this likely to happen on the current Firefox or Chrome? If not, I think we should keep things simple and just return `container.shadow_root`

ost_utils/selenium/navigation/driver.py
```diff
@@ -69,6 +71,52 @@ def find_elements(self, by, value):
...
+            return old_modal
+
+    def find_dialog_root(self, modal_id: str, immediate: bool = False) -> WebElement:
+        if immediate:
+            return self._find_dialog_root(modal_id)
+        else:
+            dialog_root = WebDriverWait(self.__driver, timeout=5).until(lambda d: self._find_dialog_root(modal_id, d))
+            return dialog_root
+
+    def find_dialog_element(self, modal_id: str, locator, waitCondition=None, message=None) -> WebElement:
```
ljelinkova, line 103: Where do we use the `find_dialog_element` with `waitCondition`?

ost_utils/selenium/navigation/driver.py
sjd78: Last time I checked, only very recent Chrome supports `shadow_root`. Firefox doesn't support it yet.

ost_utils/selenium/navigation/driver.py
sjd78: `ClusterUpgradeDialog` in the toggle functions (`toggle_check_all_hosts(self)`)

The existing functions to wait for stuff to happen don't work since they search for the document root each time so cannot see into the shadow root.

ost_utils/selenium/navigation/driver.py
ljelinkova: Ok, thanks for the explanation.

ost_utils/selenium/navigation/driver.py
```diff
@@ -69,6 +71,52 @@ def find_elements(self, by, value):
...
+        except NoSuchElementException:
+            # if not found, look for a normal modal
+            old_modal = _driver.find_element(By.ID, modal_id)
+            return old_modal
+
+    def find_dialog_root(self, modal_id: str, immediate: bool = False) -> WebElement:
+        if immediate:
+            return self._find_dialog_root(modal_id)
+        else:
+            dialog_root = WebDriverWait(self.__driver, timeout=5).until(lambda d: self._find_dialog_root(modal_id, d))
```
ljelinkova, line 100: Please try to use `wait_until` if possible here. The aim is to standardize the waits and the timeouts used throughout the tests (we use `assert_utils.SHORT_TIMEOUT` in `wait_until` and `assert_utils.LONG_TIMEOUT` in `wait_long_until`.

ost_utils/selenium/navigation/driver.py
ljelinkova: The code looks solid but inconsistent with what we have now. I would prefer the driver class offered methods with similar signature for working with both, gwt and ui-extensions. What about keeping the existing methods but adding the possibility to specify the dialog ID (meaning that we should look for shadow root)?

Something like this:
```
    def is_xpath_present(self, xpath, ui_extension_dialog_id = None):
        try:
            if ui_extension_dialog_id is None:
                self.find_element(By.XPATH, xpath)
            else self.find_dialog_element(By.XPATH, xpath)
            return True
        except NoSuchElementException:
            return False
```

Maybe you could alter the `find_element` method itself to support both, gwt and ui-extensions.  That way all the derived methods could stay untouched except for forwarding the optional dialog id argument (I am OK if you'd modified only necessary methods like this, so basically just `xpath_wait_and_click` I guess.

ost_utils/selenium/navigation/driver.py
sjd78: I tried using `wait_until` a few times with unreliable results, but I can try it out again.

I was wary of using it given the line in selenium docs (at https://www.selenium.dev/documentation/webdriver/waits/#explicit-wait) about using `WebDriveWait`:
> The first and only argument that is passed to our condition is always a reference to our driver object, WebDriver. In a multi-threaded environment, you should be careful to operate on the driver reference passed in to the condition rather than the reference to the driver in the outer scope.

It is possible I don't understand enough Python to see that `ConditionClass` is handling things ok, but it got it working reliably with direct `WebDriverWait` use.

sjd78: I can easily swap out the timeout to be the constant values for `SHORT_TIMEOUT`.  Maybe that will be enough?

ost_utils/selenium/navigation/driver.py
sjd78: @ljelinkova - I updated the code to push most of the work back in to `driver.find_element` and `drive.find_elements`.  So the users of the driver code should be more consistent now.

sjd78:
@ljelinkova - I updated the code to push most of the work back in to `driver.find_element` and `drive.find_elements`. So the users of the driver code should be more consistent now.

sjd78:
@ljelinkova , @sgratch , @michalskrivanek , @tinez - Anyone still available to look at this PR?

ost_utils/selenium/navigation/driver.py
sgratch: The `find_dialog_element` function is only used internally now and therefore might be renamed to using prefix by an underscore, i.e.  `_find_dialog_element`.

ost_utils/selenium/navigation/driver.py
```diff
@@ -70,6 +95,54 @@ def find_elements(self, by, value):
...
+        try:
+            # prefer the shadowDom version
+            shadow_host = _driver.find_element(By.ID, f'shadow-root-container-{modal_id}')
+            return self._access_shadow_root(shadow_host)
+        except NoSuchElementException:
+            # if not found, look for a normal modal
+            old_modal = _driver.find_element(By.ID, modal_id)
+            return old_modal
+
+    def find_dialog_root(self, modal_id: str, immediate: bool = False) -> WebElement:
```
sgratch, line 120: The `find_dialog_root` function is only used internally now and therefore might be renamed to using a prefix by an underscore (since `_find_dialog_root` already exists so maybe unified both or easier to rename to something reflecting the immediate mode ...).

sgratch (commented):
@sjd78 LGTM except 2 minor syntax hints improvements

ost_utils/selenium/navigation/driver.py
sjd78: Makes sense. I'll tweak things a bit.

sgratch (approved):

ost_utils/selenium/navigation/driver.py
```diff
@@ -70,6 +90,37 @@ def find_elements(self, by, value):
...
+            return container.shadow_root
+        except AssertionError:
+            # If the WebDriver's browser in use does NOT support `shadow_root`, then we need
+            # to fallback to using JavaScript.  The div containing the component will provide
+            # a valid WebElement that can be used just like the normal `shadow_root`.
+            return self.__driver.execute_script(
+                "return arguments[0].shadowRoot.querySelector('.ui-extensions-plugin-root')", container
+            )
+
+    def _find_dialog_root_now(self, modal_id: str, driver: WebDriver = None) -> WebElement:
```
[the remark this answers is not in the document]
sjd78: Sure, I'll make that change to keep things consistent within the driver.

tinez:
The failure is not necessarily caused by the PR, let's try again

sjd78:
OST passed locally for me with all of the most current images.  Trying again...

sgratch:
OST (finally) passed manually: https://rhv-devops-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/ds-ost-baremetal_manual/60016

tinez (approved):

PR #285, opened 2022-10-03
selenium: fix grafana changes on el9stream

michalskrivanek:
el9strema Grafana changed the case of Test to test, it's enough to match
just the "Save & " part, that works on both el8 and el9


tinez (approved):

tinez:
Still failing on el9stream

michalskrivanek:
error checks still need to be reimplemented

michalskrivanek:
tested on el8stream and el9stream

PR #289, opened 2022-10-06
create admin2 and use it for UI tests

michalskrivanek:
sharing REST API session and selenium session for the same admin user is
problematic as it keeps reviving the authorization even after UI logs
out. Let's use a separate user that we do NOT use for REST API calls
elsewhere


basic-suite-master/test-scenarios/test_002_bootstrap.py
```diff
@@ -1681,68 +1683,113 @@ def test_add_nonadmin_user(
         )
 
         assert keycloak.resolve_user_id(engine_api=engine_api, username=nonadmin_username) is not None
+
+        # create another admin for selenium tests later on
+        keycloak.activate_user(
```
tinez, line 1688: Are we activating the same user twice?

basic-suite-master/test-scenarios/test_002_bootstrap.py
```diff
@@ -1681,68 +1683,111 @@ def test_add_nonadmin_user(
...
                 --password-valid-to='2125-08-15 10:30:00Z' \
                 --password=pass:{nonadmin_password}"
         )
-        users_service = engine_api.system_service().users_service()
         with engine_utils.wait_for_event(engine_api.system_service(), 149):  # USER_ADD(149)
             users_service.add(
                 types.User(user_name=f'{nonadmin_username}@internal-authz', domain=engine_user_domain),
             )
+        # add another admin user as well
+        ansible_engine.shell(f"ovirt-aaa-jdbc-tool user add {engine_username}")
```
tinez, line 1716: The new codeblock looks the same (apart from the username) as the one above - let's put it in a loop

basic-suite-master/test-scenarios/test_002_bootstrap.py
```diff
@@ -1681,68 +1683,111 @@ def test_add_nonadmin_user(
...
         permissions_service.add(
             types.Permission(
-                user=nonadmin_user,
+                user=get_user(keycloak_enabled, engine_api, ansible_engine, nonadmin_username, engine_user_domain),
                 role=types.Role(
                     name='UserRole',
                 ),
             ),
         )
+    # add second admin to DC
```
tinez, line 1767: This code block is a duplicate too, let's put it in a loop

basic-suite-master/test-scenarios/test_002_bootstrap.py
michalskrivanek: makes sense ,can you post it? it's too much python for me:)

michalskrivanek:
not helping....

PR #290, opened 2022-10-07
ui: move web-ui portal test to the beginning

michalskrivanek:
logout from webadmin and logging back to web-ui under a different user
does not work correctly with keycloak, credentials/identity is being
cached by selenium and we can't figure out how to flush that...let's
just swap the order of tests so it doesn't happen in the first place.


michalskrivanek:
rhel8 passes
el8stream passes
el9stream will need additional grafana change

michalskrivanek:
grafana tests still need to be fixed for el9stream....

PR #292, opened 2022-10-13
Fix selenium timeouts

tinez:
From time to time Selenium server hangs indefinitely on http requests
made by client. Looking at the logs it seems that the netty code sees
the http request, but somehow it never reaches the code that handles the
response.

The client side blocks until it gets a response from the server.
The timeout for this path seems to be 5 mins. On the server side
however, the timeout for an inactive session to be killed is set
to 3 mins by default. That combination creates an undesired effect -
before the client realizes that the http request may have been stuck for
too long, the server already kills its session.

This patch shortens the client timeout for waiting on http
responses to 30 secs. Unfortunately, with current client implementation
in Selenium library, it's not possible to change this value
through i.e. an '__init__' argument - it rather enforces subclassing.
Although ugly, changing the 'RemoteConnection._timeout' class member
directly works for us.

Furthermore, the recent client libraries does not seem to translate an
urllib3's 'ReadTimeoutError' exception to Selenium's 'TimeoutException'
This is why we also add it to a list of handled/ignored exceptions
in a couple of places.

### DIFF
--- a/basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
+++ b/basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
@@ -18,6 +18,7 @@ import time
 import ovirtsdk4.types as types
 import pytest
 import requests
+import selenium.webdriver.remote.remote_connection
 
 from selenium import webdriver
 from selenium.webdriver.common.desired_capabilities import DesiredCapabilities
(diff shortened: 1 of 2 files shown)


basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
```diff
@@ -49,6 +50,11 @@
 LOGGER = logging.getLogger(__name__)
 
 
+# The default timeout is 5 mins. Ugly, but there's no way to change this value
+# other than subclassing.
```
ljelinkova, line 54: Please update the comment with the information that the timeout is in seconds and why we need to change that (I know you described it in the PR but lets put the info to the code too)

ost_utils/selenium/navigation/driver.py
```diff
@@ -232,16 +233,15 @@ def __call__(self, driver):
         # ignore StaleElementReferenceException and try again
         except StaleElementReferenceException:
             should_run_again = True
-        # ignore TimeoutException if caused by timeout in java
-        except TimeoutException as e:
+        # ignore TimeoutException if caused by timeout in java.
+        # ReadTimeoutError happens from time to time when netty on server side
+        # stops responding to requests.
+        except (TimeoutException, ReadTimeoutError) as e:
```
ljelinkova, line 239: I wouldn't change the handling of the `TimeoutException`- we want to retry only one very specific kind of that exception, not any exception caused e.g. by the fact that the underlying wait failed because it cannot find an element on the page. 

You could add `ReadTimeoutError` to `StaleElementReferenceException` except or if you want to log the error please create separate except for the `ReadTimeoutError`.

basic-suite-master/test-scenarios/test_100_basic_ui_sanity.py
tinez: Ack

ost_utils/selenium/navigation/driver.py
tinez: Ack.

ljelinkova (approved):

PR #293, opened 2022-10-19
Add separate machine basic suite master

didib:


didib:
Not ready for merging yet

PR #307, opened 2022-12-08
ost.sh: Add lock command

didib:
Change-Id: I680ed8a182f739a7f86b15e077ecaf77047c2802

### DIFF
--- a/lagofy.sh
+++ b/lagofy.sh
@@ -51,6 +51,15 @@ ost_status() {
 }
 
 ost_destroy() {
+    if [ -r "${OST_DEPLOYMENT_LOCK}" ]; then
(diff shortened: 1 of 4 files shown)


lagofy.sh
```diff
@@ -65,6 +74,14 @@ ost_destroy() {
     unset OST_INITIALIZED $(env | grep ^OST_IMAGES_ | cut -d= -f1)
 }
 
+ost_lock() {
+    if [ -z "$*" ]; then
+        echo "Usage: ./ost.sh lock Some reason"
+        exit 1
```
michalskrivanek, line 80: these are sourced functions, use return. 
code style is to use single line for simple exits...so just e.g. [[ -n "$*" ]] || { echo "xxx"; return 1; }

michalskrivanek (approved):
lgtm

PR #311, opened 2023-01-25
setup: remove ansible handler

michalskrivanek:
somehow this doesn't work in kickstart environment on el9. We do not
really need it for anything but a message so let's just drop it


michalskrivanek:
verified in ks environment

PR #312, opened 2023-01-27
basic: adjust memory sizes for VMs

michalskrivanek:
we seem to have frequent memory allocation issues with UEFI during
its initialization. Let's give VMs more memory without really increasing
guaranteed mem, so UEFI can take whatever it needs and it will get
ballooned anyway if the host doesn't have enough later on.

### DIFF
--- a/basic-suite-master/test-scenarios/test_002_bootstrap.py
+++ b/basic-suite-master/test-scenarios/test_002_bootstrap.py
@@ -1271,17 +1271,15 @@ def test_add_blank_vms(engine_api, ost_cluster_name):
     )
 
     vm_params.name = BACKUP_VM_NAME
-    vm_params.memory = 160 * MB
-    vm_params.memory_policy.guaranteed = 106 * MB
+    vm_params.memory = 512 * MB
+    vm_params.memory_policy.guaranteed = 160 * MB
     vms_service.add(vm_params)
     backup_vm_service = test_utils.get_vm_service(engine, BACKUP_VM_NAME)
 
     vm_params.name = VM0_NAME
-    least_hotplug_increment = 256 * MB
-    required_memory = 160 * MB
-    vm_params.memory = required_memory
-    vm_params.memory_policy.guaranteed = required_memory
-    vm_params.memory_policy.max = required_memory + least_hotplug_increment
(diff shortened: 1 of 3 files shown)


basic-suite-master/test-scenarios/test_002_bootstrap.py
```diff
@@ -1271,17 +1271,15 @@ def test_add_blank_vms(engine_api, ost_cluster_name):
     )
 
     vm_params.name = BACKUP_VM_NAME
-    vm_params.memory = 160 * MB
-    vm_params.memory_policy.guaranteed = 106 * MB
```
michalskrivanek, line 1275: apparently a typo from previous patch. i don't think it ever ballooned so much but better to align on the intended 160

basic-suite-master/test-scenarios/test_002_bootstrap.py
```diff
@@ -1347,10 +1345,10 @@ def test_add_blank_high_perf_vm2(engine_api, ost_dc_name, ost_cluster_name):
             ),
             memory_policy=sdk4.types.MemoryPolicy(
                 ballooning=False,
-                guaranteed=106 * MB,
+                guaranteed=160 * MB,
```
michalskrivanek, line 1348: ...at least I *think* it was a typo.

basic-suite-master/test-scenarios/test_004_basic_sanity.py
```diff
@@ -926,8 +926,8 @@ def test_add_vm1_from_template(engine_api, cirros_image_template_name, ost_clust
             use_latest_template_version=True,
             stateless=True,
             memory_policy=types.MemoryPolicy(
-                guaranteed=vm_memory,  # with so little memory we don't want guaranteed to be any lower
-                ballooning=False,
+                guaranteed=160 * MB,
+                ballooning=True,
```
michalskrivanek, line 930: this changes behavior, but with bigger guaranteed I don't think it's a concern anymore

tinez (approved):

michalskrivanek:
doesn't help. but it shouldn't really hurt either. and if we update cirros to 6.x it also seem to have higher memory requirements

PR #314, opened 2023-01-30
requirements.txt: use specific openstacksdk commit 

erav:
Openstacksdk>=0.62.0 is now published on pypi. Use it going forward.

Change-Id: I26a094b522202ace8c5e0d30749571a46d5d25d9

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ requests
 selenium==4.4.3
 # network suite deps
 # TODO Use pip version once released openstacksdk>=0.62.0
-git+https://github.com/openstack/openstacksdk.git@master
+git+https://github.com/openstack/openstacksdk.git@44c5515
 # ost_utils
 paramiko
 PyYAML


almusil:
I don't think that this will work. The message states `<=0.98.999` latest release is `0.103.0` and the comments says that we need `>=0.62.0`, closest to that is `0.99.0` which is out of range already. So I guess we will need to pin to some commit in openstacksdk before `0.99.0` release or use newer openstack ansible that is compatible with those newer versions (this step would probably require rewrite of the used roles).

mwperina:
Isn't there the correct packaged RPM we can use?

erav:
> Isn't there the correct packaged RPM we can use?

that's what i thought would happen automatically when I changed to `openstacksdk`. So Ales, do we have an el8s vs. openstack-ansible compatibility issue?

almusil:
I don't think there is a rpm, openstack did only release of `0.99.0` after `0.61.0`. It's not about el8 or el9 this is taken from pip.

erav:
> I don't think there is a rpm, openstack did only release of `0.99.0` after `0.61.0`. It's not about el8 or el9 this is taken from pip.

and where is the version of openstack-ansible decided? by the os version of ansible? if so, pinning to a pre-0.99 might break el9?

almusil:
That's decided by the version installed, pinning doesn't help because only version available before `0.99.0` is `0.61.0` which would break the tests.

almusil:
another solution might be to pin specific commit from openstacksdk (not sure if it's possible with pip). We need [0], so pinning that one should work without breaking anything else.

[0] https://github.com/openstack/openstacksdk/commit/44c551560242379daa3ae3571aa260f7c6858cb7

erav:
ok I'll try

almusil (approved):

michalskrivanek:
good enough

PR #316, opened 2023-01-31
basic: add sleep for cluster upgrade dialog

michalskrivanek:
let's try to wait a bit, seems the regular waiting on dialog is not
enough


ljelinkova (approved):

michalskrivanek:
one run is not enough but let's go ahead and see....

PR #317, opened 2023-03-13
ansible: Start image_template from engine

mnecas:
The image_template did not start the playbook from engine but from some
diffrent place. With this patch the playbook is copied to the engine and
started there.


michalskrivanek:
doesn't make any sense....

michalskrivanek:
...but that's ok, it's black magi^H^Hansible

michalskrivanek (approved):

PR #320, opened 2023-11-23
selenium: Create artifacts dir manually

tinez:
OST started complaining that `{selenium_remote_artifacts_dir}` doesn't exist. We never created it by ourselves, so it's probable some new podman version changed behavior in this regard.


michalskrivanek:
failure is later in console.vv donwload only, so let's go with this...

PR #323, opened 2023-11-24
selenium: Change permissions on artifacts dir

tinez:
The directory is created with root:root and 755 permissions. Chrome container runs under 'seluser', so it's lacking write permissions.


michalskrivanek:
it got to grafana!:)

michalskrivanek:
which should be fixed by https://github.com/oVirt/ovirt-dwh/pull/63

PR #329, opened 2024-08-15
Rename ost-of-ost.yaml to ost-of-ost.disabled

michalskrivanek:
disabling the trigger as the infrastructure to run it on no longer available


michalskrivanek:
trying out if renaming is good enough. this affects only this project, the trigger needs to be disabled from all the others too

michalskrivanek:
yeah, works good enough. No more triggers, "/ost" commands are disabled

PR #344, opened 2025-11-06
he-basic-suite-master: deploy order

dupondje:
Define the order of he_deploy.
Cause without strict order, the test_run_dig_loop did run before the test_he_deploy has even been executed.


dupondje:
Not needed anymore.

PR #350, opened 2026-06-19
system-tests: add NVMe-oF/TCP storage domain tests

tdemaret:
Add NVMe-oF/TCP storage domain support to oVirt. This is a block/shared storage domain type, analogous to iSCSI, using the built-in Linux NVMe 
initiator (nvme-cli). It supports both authenticated (DH-HMAC-CHAP) and non-authenticated NVMe-oF targets.

The feature try to be backend-agnostic for any NVMe/TCP target works (Ceph NVMe-oF Gateway is the reference implementation).

# Change :
- New storage domain type: nvmeof in REST API / Ansible / SDK
- Authentication : optional DH-HMAC-CHAP 
- Multipath : native NVMe multipath detected automatically; falls back to dm-mpath
- Host deployment: NVMe host NQN auto-detected from /etc/nvme/hostnqn

# Limitations :
- No iSCSI bond support for NVMe-oF
- NVMe-oF storage domain cannot be edited or imported via webadmin

### DIFF
--- a/ansible-suite-master/test-scenarios/test_002_ansible.py
+++ b/ansible-suite-master/test-scenarios/test_002_ansible.py
@@ -114,7 +114,16 @@ def test_ansible_run(
                     "lun_id": lun.get_uuids(sd_iscsi_ansible_host)[:2],
                 }
             },
+            "nvmeof": {
+                "nvmeof": {
+                    "nqn": lun.get_nvmeof_connection_info(sd_iscsi_ansible_host)["nqn"],
+                    "port": 4420,
+                    "address": sd_iscsi_host_ip,
+                    "lun_id": lun.get_nvmeof_uuids(sd_iscsi_ansible_host),
+                }
+            },
         },
+
         logical_networks=[
             {
                 "name": "Migration_Net",
--- a/basic-suite-master/test-scenarios/conftest.py
+++ b/basic-suite-master/test-scenarios/conftest.py
@@ -87,3 +87,13 @@ def sd_iscsi_ansible_host(
     ansible_storage,
 ):  # pylint: disable=function-redefined
     return ansible_storage
+
+
+@pytest.fixture(scope="session")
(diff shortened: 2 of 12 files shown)


peter-boden (changes requested):
Three gaps that I think you should cover:                                                                                                                                                                                 
                                                                                                                                                                                                                                         
 1. Multipathing. The setup provisions one subsystem with one namespace. All CI distros have native NVMe multipath on by default (CONFIG_NVME_MULTIPATH=y), so the native enumeration branch in vdsm does run, but with a single path.   
 A second controller (or a second subsystem backed by the same device) would give real multipath coverage and also exercise the     
 connection-dedup path.                                                                                                                                                                                                                  
                                                                                                                                                                                                                                         
 2. Authentication. The setup disables auth entirely (allow_any_host=1 in setup_storage.sh:222). DH-HMAC-CHAP is cheap to add: nvme gen-dhchap-key ships in the nvme-cli version this feature needs, and kernel support                  
 (CONFIG_NVME_TARGET_AUTH) is standard on RHEL 9. This is worth adding now because auth is currently broken in the companion vdsm PR (https://github.com/oVirt/vdsm/pull/481) (wrong nvme connect flags: -w/-k instead of -q/-S), and an 
 auth test would give a concrete red/green signal for that fix.                                                                                                                                                                                                       
                                                                                                                                                                                                                                         
 3. NVMe as a secondary storage domain. NVMe-oF is currently an alternate master storage domain, gated behind OST_MASTER_SD_TYPE=nvmeof, so a CI run tests iSCSI or NVMe as master, never both. The target is already provisioned on     
 every basic-suite run (sdc is the NVMe-oF device, setup_nvmeof runs unconditionally at setup_storage.sh:294), so adding NVMe as a secondary storage domain in the default NFS-master run would be a small change (add                   
 add_nvmeof_storage_domain to the VectorThread in test_add_secondary_storage_domains) and would let us test both storage types in one go.

PR #353, opened 2026-09-08
Linting

dupondje:


basic-suite-master/test-scenarios/test_002_bootstrap.py
```diff
@@ -4,44 +4,34 @@
 #
 # -*- coding: utf-8 -*-
```
JasperB-TeamBlue: the coding utf header is dropped in `basic-suite-master/test-scenarios/conftest.py` but not here, why?

network-suite-master/fixtures/host.py
```diff
@@ -52,7 +51,7 @@ def _non_spm_host(system, host_ids):
     for host_id in host_ids:
         host = hostlib.Host(system)
         host.import_by_id(host_id)
-        if host.is_not_spm or id == host_ids[-1]:
+        if not host.is_not_spm or host_id == host_ids[-1]:
```
peter-boden: I think this `not` doesn't need to be there? 
You go from "host is not spm or ..." to "host is spm or ..."?

peter-boden (changes requested):
One issue I think, otherwise LGTM.

common/test-scenarios-files/vds_hooks/add_mdevs.py
```diff
@@ -6,57 +6,59 @@
...
+                "nvidia-14": {
+                    "name": "GRID M10-5C",
+                    "available_instances": 16,
+                    "description": "num_heads=2, frl_config=45, framebuffer=2048M, max_resolution=4096x2160, max_instance=16",
                 },
-		"nvidia-14" : {
-                    "name" : "GRID M10-5C",
-                    "available_instances" : 16,
-                    "description" : "num_heads=2, frl_config=45, framebuffer=2048M, max_resolution=4096x2160, max_instance=16"
+                "nvidia-21": {
```
JasperB-TeamBlue, line 41: What does bumping these values have to do with linting?

network-suite-master/ovirtlib/netattachlib.py
```diff
@@ -180,8 +180,10 @@ def to_network_attachments(network_attachments_data):
         return [attachment.to_network_attachment() for attachment in network_attachments_data]
 
 
-class BondingData(object):
-    def __init__(self, name, slave_names, options={}):
+class BondingData:
+    def __init__(self, name, slave_names, options=None):
```
JasperB-TeamBlue, line 184: what is the use of first setting the default to None to then set it to an empty dict instead of directly setting it to an empty dict?

ost_utils/backend/base.py
```diff
@@ -130,45 +130,46 @@ def macs_for(self, hostname, network_name):
...
+    @cached_property
     def bonding_network_name(self):
-        return next(nn for nn in self.network_names() if "bonding" in nn)
+        return next(nn for nn in self.network_names if "bonding" in nn)
+
+    @cached_property
+    def _management_network_ips(self):
+        return next(iter(self.ip_mapping().values()))[self.management_network_name]
 
-    @cache
```
JasperB-TeamBlue, line 167: why does this cache tag get removed and not replaced like with the other items?

JasperB-TeamBlue (changes requested):
Some questions, remarks. Otherwise looks good!

basic-suite-master/test-scenarios/test_002_bootstrap.py
dupondje: In fact it was useless, because not at top of the file.
Removed all the ones left.

common/test-scenarios-files/vds_hooks/add_mdevs.py
dupondje: The diff is a bit weird on github.
The only change here is it removes the space between "key" and :

network-suite-master/ovirtlib/netattachlib.py
dupondje: https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

ost_utils/backend/base.py
dupondje: Because it has an argument, the cache_property cannot be used

common/test-scenarios-files/vds_hooks/add_mdevs.py
dupondje: The diff is shown weirdly on github. I only removed the space between key and :

peter-boden (approved):

JasperB-TeamBlue (approved):
LGTM!

